PR #24, opened 2014-09-22
Added support for slave servers

cbergmann:
Hi,

this change is the only reason why I can't use the upstream version.
Would it be possible to include it into theforeman/puppet-dns?
Thanks in advance.
Clemens


mmoll:
continuation in #25

PR #25, opened 2014-10-14
Slave support

treydock:
This is an extension to #24 

Commit message:

    Add support for master-slave configuration

    This also changes the use of defining the "dns" class inside the dns::zone defined type.
    Instead the dns class is included and the parameters "zonefilepath" and "namedservicename" pull their values from
    the dns class if left as undef.

    Initial work came from PR #24.  This varies from that pull request in the following ways

    * allow-transfer is not inside the "if master" condition
    * A few syntax errors corrected
    * Use "join" to generate values for allow-transfer and masters

Also wanted to add I tried to keep my changes that do not relate to functionality of master-slave to a minimum.  Defining parameter defaults in `dns::zone` that came from `dns::params` was causing problems with unit tests without setting a pre_condition.  Rather than use a pre_condition in the unit tests I removed the `dns` class declaration from `dns::zone` and instead us the safer `include`.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,7 @@
 fixtures:
   repositories:
     concat_native:    'https://github.com/theforeman/puppet-concat'
+    stdlib:           'https://github.com/puppetlabs/puppetlabs-stdlib.git'
 
   symlinks:
     dns: "#{source_dir}"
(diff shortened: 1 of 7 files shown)


mmoll:
@ekohl any results? :)

ekohl:
Now that was a long weekend ... If only I hadn't forgotten about it :(

domcleal:
Merged as part of #28.  Big thanks to @treydock and @cbergmann for the new functionality and tidyup.

PR #29, opened 2015-03-07
remove check if SOA record is in the same zone

mmoll:
I can't find anything in RFCs for that requirement and am even running DNS servers myself with zones that have SOAs in other zones.


domcleal:
Thanks @mmoll!

PR #33, opened 2015-05-04
Add support for managing static records

bennycornelissen:
Currently, this module is incapable of managing any records within the zones it creates. Partly, this makes sense because Foreman would be managing the A and PTR records for the hosts it creates. However, users might want to manage other records (A + PTR records for secondary interfaces, CNAME records, TXT records, etcetera), that cannot be easily managed through Foreman. 

I have added functionality to make the above possible by introducing a `dns::record` type, and adding a second file for each zone (db.myzone.tld.static). The `dns::record` type creates a concat fragment that is later collected by the `dns::zone` type. Exported resources are also collected there. 

If you don't want to use this functionality, then the only change will be that the module creates an additional file for each zone: db.myzone.tld.static, which will be empty. Foreman will still manage the db.myzone.tld file. 

Effectively this allows you to manage CNAME records for servers by including them in the Hiera data for the specific host. For example, let's implement the DNS module in a profile: 

```
# 
class profile::dnsserver {
  include dns

  create_resources( 'dns::zone', hiera('dns::zones', {} ))
  create_resources( 'dns::record', hiera('dns::records', {} ))

}
```
and add something to a base profile so that CNAME records from Hiera data (for servers that are not DNS servers) are exported: 
```
class linuxbase::dns_record {

  # collect dns records from hieradata and create exported resources
  $dns_record_defaults = {
    comment => "Created by ${::fqdn}"
  }
  if ! defined( Class['dns'] ) {
    create_resources('@@dns::record', hiera('dns::records', {} ), $dns_record_defaults )
  }

}
```

Then add some Hiera data for the DNS server:
```
---
classes: 
  - 'profile::dnsserver'

dns::zones:
  'example.com':
    zonetype: 'master'
    allow_transfer:
      - 172.16.32.13
    reverse: false
  '8.9.10.in-addr.arpa':
    zonetype: 'master'
    allow_transfer:
      - 172.16.32.13
    reverse: true
```
And for a non-DNS server:
```
---
dns::records:
  'icanhazwebsite.example.com':
    target: 'vm124.example.com.'
    type: 'CNAME'
```

On the DNS server, you will find a file called `db.example.com.static` with the following contents: 
```
..
icanhazwebsite  IN  CNAME  vm124.example.com.  ; Created by vm124.example.com
..
```

The `dns::record` type currently supports the following DNS record types: 
- A
- PTR
- CNAME
- TXT
- SRV
- MX

### DIFF
--- a/manifests/record.pp
+++ b/manifests/record.pp
@@ -1,0 +1,61 @@
+# Define new DNS record
+define dns::record (
+  $zone     = undef,
+  $label    = undef,
+  $target   = undef,
+  $type     = undef,
+  $priority = undef,
+  $weight   = undef,
+  $port     = undef,
+  $comment  = ' ',
+) {
+
+  if $target == undef {
+    fail("Failed to define target for record ${title}")
+  }
+  else {
+    validate_string($target)
+  }
+
+  if $type == undef {
+    fail("Failed to define record type for record ${title}")
+  }
+  else {
+    validate_string($type)
+    $dnstype = upcase($type)
+  }
+
+  # if you choose not to override the label or zone attributes  (if you use the
+  # FQDN of the record you want to create, you shouldn't need to), we will 
+  # generate these attributes from the FQDN.
+  if ($label == undef) or ($zone == undef) {
+    $domain_array = split($title, '[.]' )
+    $_label = $domain_array[0]
+    $_zone = join(delete_at($domain_array, 0), '.')
+  } else {
+    $_label = $label
+    $_zone = $zone
+  }
+  
+  if $dnstype =~ /(CNAME|TXT|A|PTR)/ {
+    $line = "${_label} IN ${dnstype} ${target}  ; ${comment}"
+  }
+  elsif $dnstype == 'MX' {
+    validate_integer($priority, 65535, 0)
+    $line = "${_label} IN ${dnstype} ${priority} ${target}  ; ${comment}"
+  }
+  elsif $dnstype == 'SRV' {
+    validate_integer($priority, 65535, 0)
+    validate_integer($weight, 65535, 0)
+    validate_integer($port, 65535, 0)
+    $line = "${_label} IN ${dnstype} ${priority} ${weight} ${port} ${target}  ; ${comment}"
+  }
+  else {
+    fail("Incorrect DNS record type specified for record ${title}")
+  }
+
+  concat_fragment { "dns-static-${_zone}+02content-${title}.dnsstatic":
+    content => $line,
+  }
+
+}
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -16,9 +16,10 @@ define dns::zone (
     $contact        = "root.${title}.",
     $zonefilepath   = $::dns::zonefilepath,
     $filename       = "db.${title}",
+    $static_records = false,
 ) {
 
-  validate_bool($reverse)
+  validate_bool($reverse, $static_records)
   validate_array($masters, $allow_transfer)
 
   $zonefilename = "${zonefilepath}/${filename}"
@@ -36,4 +37,31 @@ define dns::zone (
     replace => false,
     notify  => Service[$::dns::namedservicename],
   }
+
+  if $static_records {
+    concat_fragment { "dns-static-${zone}+01header.dnsstatic":
+      content => "; static file for ${zone}",
+    }
+  
+    Dns::Record <<| |>> { notify => Concat_build["dns-static-${zone}"] }
+  
+    concat_build { "dns-static-${zone}":
+      order  => [ '*.dnsstatic' ],
+      notify => [
+        Service[$::dns::namedservicename],
+        File[ "${zonefilename}.static"],
+      ],
+    }
+  
+    file { "${zonefilename}.static":
+      ensure  => file,
+      owner   => $dns::user,
+      group   => $dns::group,
+      mode    => '0644',
+      notify  => Service[$::dns::namedservicename],
+      source  => concat_output("dns-static-${zone}"),
+      require => Concat_build[ "dns-static-${zone}"],
+    }
+  }
+
 }
--- a/spec/defines/dns_record_spec.rb
+++ b/spec/defines/dns_record_spec.rb
@@ -60,4 +60,20 @@ describe 'dns::record' do
     it { should_not compile }
   end
 
+  context 'when creating SRV record with non-integer weight' do
+    let(:title) { "badsrv.example.com" }
+    let(:params) {{ 
+      :label  => "_http._tcp.badsrv.example.com.",
+      :target => "vm123.example.com.",
+      :zone   => "example.com",
+      :priority => "10",
+      :weight   => "badweight",
+      :port     => "80",
+      :type   => "SRV",
+      :comment => "Created by vm123",
+    }}
+    
+    it { should_not compile }
+  end
+
 end
--- a/spec/defines/dns_record_spec.rb
+++ b/spec/defines/dns_record_spec.rb
@@ -1,0 +1,63 @@
+require 'spec_helper'
+
+describe 'dns::record' do
+
+  let(:facts) do
+    {
+      :osfamily   => 'RedHat',
+      :fqdn       => 'puppetmaster.example.com',
+      :clientcert => 'puppetmaster.example.com',
+      :ipaddress  => '192.168.1.1'
+    }
+  end
+
+  let :pre_condition do
+    'include dns'
+  end
+
+  context 'when creating CNAME record' do
+    let(:title) { "foo.example.com" }
+    let(:params) {{ 
+      :target => "vm123.example.com.",
+      :type   => "CNAME",
+      :comment => "Created by vm123",
+    }}
+  
+    it "should have valid record configuration" do
+      verify_concat_fragment_exact_contents(subject, 'dns-static-example.com+02content-foo.example.com.dnsstatic', [
+        'foo IN CNAME vm123.example.com.  ; Created by vm123',
+      ])
+    end
+  end
+
+  context 'when creating SRV record' do
+    let(:title) { "www.example.com-http-backend1" }
+    let(:params) {{ 
+      :target => "vm123.example.com.",
+      :label  => "_http._tcp.www.example.com.",
+      :zone   => "example.com",
+      :priority => "10",
+      :weight   => "60",
+      :port     => "80",
+      :type   => "SRV",
+      :comment => "Created by vm123",
+    }}
+  
+    it "should have valid record configuration" do
+      verify_concat_fragment_exact_contents(subject, 'dns-static-example.com+02content-www.example.com-http-backend1.dnsstatic', [
+        '_http._tcp.www.example.com. IN SRV 10 60 80 vm123.example.com.  ; Created by vm123',
+      ])
+    end
+  end
+
+  context 'when creating CNAME with blank target' do
+    let(:title) { "badcname.example.com" }
+    let(:params) {{ 
+      :type   => "CNAME",
+      :comment => "Created by vm123",
+    }}
+    
+    it { should_not compile }
+  end
+
+end
--- a/spec/defines/dns_zone_spec.rb
+++ b/spec/defines/dns_zone_spec.rb
@@ -54,6 +54,34 @@ describe 'dns::zone' do
     ])
   end
 
+  context 'when including static records' do
+    let(:params) {{ :static_records => true }}
+    it "should create zone static file" do
+      should contain_file('/var/named/dynamic/db.example.com.static').with({
+        :owner    => 'named',
+        :group    => 'named',
+        :mode     => '0644',
+        :notify   => 'Service[named]',
+      })
+    end
+ 
+    it "should have valid zone file contents" do
+      verify_exact_contents(subject, '/var/named/dynamic/db.example.com', [
+        '$TTL 10800',
+        '@ IN SOA puppetmaster.example.com. root.example.com. (',
+        '	1	;Serial',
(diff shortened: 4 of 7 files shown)


templates/zone.header.erb
```diff
@@ -11,3 +11,5 @@ $TTL <%= @ttl %>
 <% if ! @reverse %>
 <%= @soa %>. IN A <%= @soaip %>
 <% end %>
+
+$INCLUDE <%= @zonefilepath -%>/<%= @filename -%>.static <%= @zone -%>.
```
ekohl: I'd prefer if this was hidden under a boolean so most users wouldn't have this enabled.

spec/defines/dns_record_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  it "should have valid record configuration" do
+    verify_concat_fragment_exact_contents(subject, 'dns-static-example.com+foo.example.com.dnsstatic', [
+      'foo IN CNAME vm123.example.com.  ; Created by vm123',
+    ])
+  end
+
+end
+
+at_exit { RSpec::Puppet::Coverage.report! }
```
ekohl: I don't think this is needed here. Maybe a nice addition to the spec helper.

templates/zone.header.erb
bennycornelissen: I introduced a `static_records` variable in the `dns::zone` type, which defaults to false. Rspec testing tests both implementations (static_records true as well as false)

spec/defines/dns_record_spec.rb
bennycornelissen: You are right. Removed the line.

manifests/record.pp
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    $_label = $label
+    $_zone = $zone
+  }
+  
+  if $dnstype =~ /(CNAME|SRV|TXT|A|PTR)/ {
+    $line = "${_label} IN ${dnstype} ${target}  ; ${comment}"
+  }
+  elsif $dnstype == 'MX' {
+    $line = "${_label} IN ${dnstype} ${priority} ${target}  ; ${comment}"
```
ekohl: I wonder why you special case MX here. SRV also has a priority.

manifests/zone.pp
```diff
@@ -36,4 +38,34 @@
     replace => false,
     notify  => Service[$::dns::namedservicename],
   }
+
+  if $static_records == true {
+    concat_fragment { "dns-static-${zone}+01header.dnsstatic":
+      content => "; static file for ${zone}",
+    }
+  
+    Dns::Record <<| |>> { notify => Concat_build["dns-static-${zone}"] }
```
ekohl: I don't think this is needed since concat_fragment does this by itself.

manifests/record.pp
```diff
@@ -0,0 +1,38 @@
...
+  }
+  else {
+    warning('Incorrect DNS record type specified')
+  }
+
+  concat_fragment { "dns-static-${_zone}+${title}.dnsstatic":
```
ekohl: I'd include ```02content``` in there so you're sure the header is always first (00 is a valid DNS label): ```dns-static-${_zone}+02content-${title}.dnsstatic```.

manifests/record.pp
```diff
@@ -0,0 +1,38 @@
+# Define new DNS record
+define dns::record (
```
ekohl: Can you re-use $title here somehow?

ekohl:
I won't nitpick about whitespace until we hash out the rest, but I will once we get closer ;)

ekohl:
btw, I think I like `$INCLUDE` here better than the approach in https://github.com/theforeman/puppet-dns/pull/16 so :+1: for that.

manifests/zone.pp
bennycornelissen: I found that without the 'notify' part I needed a second run to get the records to show up.

manifests/record.pp
bennycornelissen: Good point, will fix.

manifests/record.pp
bennycornelissen: Agreed

manifests/zone.pp
ekohl: Interesting. @domcleal any idea?

manifests/record.pp
ekohl: Never mind, I see you do use it in the code itself.

bennycornelissen:
Was there any other whitespace you wanted removed? ;-)

spec/defines/dns_zone_spec.rb
```diff
@@ -39,6 +39,7 @@
     })
   end
 
+
```
ekohl: Nitpick: can you remove this line?

spec/defines/dns_zone_spec.rb
```diff
@@ -54,6 +55,35 @@
...
+        '	3600	;Negative caching TTL',
+        ')',
+        '@ IN NS puppetmaster.example.com.',
+        'puppetmaster.example.com. IN A 192.168.1.1',
+        '$INCLUDE /var/named/dynamic/db.example.com.static example.com.'
+      ])
+    end
+  end
+
+
```
ekohl: Nitpick: can you remove this line?

spec/defines/dns_zone_spec.rb
```diff
@@ -73,6 +103,8 @@
     end
   end
 
+
+
```
ekohl: Nitpick: can you remove these 2 lines?

spec/defines/dns_record_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  it "should have valid record configuration" do
+    verify_concat_fragment_exact_contents(subject, 'dns-static-example.com+02content-foo.example.com.dnsstatic', [
+      'foo IN CNAME vm123.example.com.  ; Created by vm123',
+    ])
+  end
+
+end
+
```
ekohl: Nitpick: can you remove this line?

manifests/zone.pp
```diff
@@ -36,4 +37,31 @@
...
+    file { "${zonefilename}.static":
+      ensure  => file,
+      owner   => $dns::user,
+      group   => $dns::group,
+      mode    => '0644',
+      notify  => Service[$::dns::namedservicename],
+      source  => concat_output("dns-static-${zone}"),
+      require => Concat_build[ "dns-static-${zone}"],
+    }
+  } # unless
```
ekohl: I think the comment doesn't make sense here.

manifests/record.pp
```diff
@@ -0,0 +1,44 @@
...
+  $zone     = undef,
+  $label    = undef,
+  $target   = undef,
+  $type     = undef,
+  $priority = undef,
+  $weight   = undef,
+  $comment  = ' ',
+)
+{
```
ekohl: Nitpick: we generally write ```) {``` in a single line.

spec/defines/dns_record_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+    }
+  end
+
+  let(:title) { "foo.example.com" }
+  let(:params) {{ 
+    :target => "vm123.example.com.",
+    :type   => "CNAME",
+    :comment => "Created by vm123",
+  }}
+
```
ekohl: Nitpick: can remove this line?

ekohl:
There, some nitpicks :) Also, I will ask to squash all commits before we merge. All in all I think this can be merged soon.

ekohl:
:+1:

mmoll:
I don't use the dns module, but :+1: visually. Please squash also the last two commits into the other.

bennycornelissen:
I will squash the last few commits, although I was planning on adding a few more test cases tonight (if I find the time), so input validation is actually tested :smile:

mmoll:
more tests are always good :green_heart:

manifests/record.pp
```diff
@@ -0,0 +1,56 @@
...
+  $type     = undef,
+  $priority = undef,
+  $weight   = undef,
+  $port     = undef,
+  $comment  = ' ',
+) {
+
+  $dnstype = upcase($type)
+
+  if $target == undef {
```
ekohl: You can use ```validate_string($target)```

manifests/record.pp
```diff
@@ -0,0 +1,56 @@
...
+    fail("No target specified for record ${title}")
+  }
+
+  if $dnstype == undef {
```
ekohl: Here validate_string works as well.

manifests/zone.pp
```diff
@@ -36,4 +37,31 @@
     replace => false,
     notify  => Service[$::dns::namedservicename],
   }
+
+  if $static_records == true {
```
ekohl: Nitpick: ```== true``` can be left out.

templates/zone.header.erb
```diff
@@ -11,3 +11,6 @@ $TTL <%= @ttl %>
 <% if ! @reverse %>
 <%= @soa %>. IN A <%= @soaip %>
 <% end %>
+<% if @static_records == true %>
```
ekohl: Nitpick: ```== true``` can be left out.

manifests/record.pp
bennycornelissen: I think I can even totally drop the check for empty $dnstype, as it's caught by the default of the if statement starting line 34 (which I should probably be replacing with a case statement by now :wink: )

bennycornelissen:
I'm going to squash this last commit and then call it a day guys. :smile:

bennycornelissen:
Hold off on merging please. I have found a really nasty bug that needs to be fixed first. 

The problem is that updates to a static zone file (db.example.com.static) will not be transferred to a slave, because updates to the *.static file do not result in an incremented serial number for the zone.  Also, whenever a slave transfers a zone for the first time, the *.static file the slave has becomes useless because it's no longer included. I have tried moving the SOA record to the *.static file, which makes it trivial to update the serial on each Puppet run, but because Foreman uses nsupdate, the SOA record's serial is not used at all. Even worse, when you restart (not reload) BIND, the INCLUDE line is removed entirely, which makes it impossible to add new static records after bind has been restarted. 

As a possible fix I am considering the following: 
1. Slave zones are no longer provisioned with any static data. They must simply get all the data in a zone transfer from the master
2. Rebuild the *.static file from a zonefile into a nsupdate file. After each change to the file, run nsupdate on the zone. This will increment the serial number and make things work.

bennycornelissen:
If I don't want to rewrite everything, there is a second solution: if you want to use static records inside a zone, you cannot use that zone for A/PTR records in Foreman. That would allow you to do something like this: 

```
dns::zone{ 'dynamic.example.com': }

dns::zone{ 'static.example.com': static_records => true, } 
```

If you create a dynamic zone, or a slave zone of any kind, the module will work like the master branch does currently. So the zone dynamic.example.com can be added in Foreman, can hold all the dynamic records Foreman wants to store there, but can **not** hold static records. The exact opposite goes for the zone static.example.com. You can **not** use it in Foreman, but you can store whatever static records you want in there. 

I'll admit, this is not quite as nice as having support for static records in any zone, regardless of whether the zone is dynamically managed or not. However, the only alternative I see for now is finding a way to use nsupdate to add/remove the static records. Adding them is easy, but I'm not quite sure how I would remove records.

bennycornelissen:
@ekohl 
Update: I'm pursuing 2 options currently, but I would like your input. 

Option 1: a zone is either a dynamic zone (so no support for static records, but works with Foreman) or a static zone (so support for static records, but doesn't work with Foreman). This works pretty well, although I'm still testing with methods to update the serial (either setting it during each run, or using an exec statement to set it when something changes). 

Option 2: finding a way of using nsupdate to add and remove static records. I have a working proof-of-concept for this, using a shellscript that gets triggered whenever a 'nsupdate' file is changed by Puppet. The shellscript currently does the following: 

- get the rndc key from config
- make a copy of the nsupdate file 
- if an earlier copy of the file exists, diff the old and the current file, and for each record that exists in the old file, but not in the current file, add a 'update delete' statement in the current file
- use the current file to run nsupdate
- replace the old file with the copy of the new file

This way, when you use exported resources to create DNS records, the records are purged automatically whenever you decomission a host. 

So basically, it works, but I think the shellscript if a fairly dirty hack.

ekohl:
I'll have to think about it for a bit. These are hard problems.

bennycornelissen:
I agree. I have built a rough draft of how things would work when using nsupdate. You can find it here: https://github.com/bennycornelissen/puppet-dns/tree/staticrecords_fix

A side effect of this method is that each zone has 4 files in the `$zonefilepath`, like this: 
```
-rw-r--r-- 1 bind bind  351 May  8 13:50 db.example.com
-rw-r--r-- 1 bind bind 2071 May  8 13:53 db.example.com.jnl
-rw-r--r-- 1 bind bind  144 May  8 13:53 db.example.com.nsupdate
-rw-r--r-- 1 root root  144 May  8 13:53 db.example.com.nsupdate.old
```
Here, `db.example.com` is the regular zonefile, containing the SOA and NS records, the `db.example.com.jnl` is the journal containing dynamic updates to the zone, which will be written to `db.example.com` whenever BIND is stopped or restarted. It also holds the one true serial BIND actually cares about. The `db.example.com.nsupdate` file replaces the `db.example.com.static` file in my 'old' implementation, and is a proper nsupdate file containing the static records. The `db.example.com.nsupdate.old` file is a backup copy that is being made whenever `nsupdate` is actually run. This way, whenever a static record is removed, the `nsupdate_wrapper` script can figure out which records to remove.

bennycornelissen:
@ekohl and myself discussed the issue further on IRC, and I'm closing this PR.

PR #34, opened 2015-05-11
Add support for managing static records using nsupdate

bennycornelissen:
As a result of the findings in #33 I have rebuilt my implementation to use nsupdate instead. This way, we manage static records the very same way Foreman manages the dynamic records of the zone, and we have no serial issues. I have chosen to use a wrapper script (nsupdate_wrapper) because it helps to handle deletion of unmanaged records (leftover exported resources after a node is decommissioned). 

In the future I want to try and re-solve this issue by creating an nsupdate provider, but for now this implementation works.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,4 @@
+--format documentation
+--color
+--tty
+--backtrace
--- a/README.md
+++ b/README.md
@@ -22,6 +22,23 @@ Zones can be created with the `dns::zone` resource:
 Slaves can also be configured by setting `allow_transfer` in the master's zone
 and setting `zonetype => 'slave'` in the slave's zone.
 
+# Static records
+
+Create 'static' DNS records, such as CNAME records, using the `dns::record` type.
+
+    dns::record { 'foo.example.com':
+      target  => 'vm123.example.com.'
+      type    => 'CNAME'
+    }
+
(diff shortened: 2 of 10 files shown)


jbcraig:
I like the direction of this change.  The ability to add static records dynamically is appealing.  One thing I noted was how you override label/zone in the records.pp.  You should test and override them separately, rather than as a unit.  There are cases where the title is fine for label but you simply need to override the zone setting.  This is most apparent in PTR records where a subnet other than a 24 bit boundary will result in erroneous zone value, but the title is fine as a label.

bennycornelissen:
The idea is that when you use something other than an FQDN or fully qualified pointer for the title, then you need so set both label and domain. If you omit one, it should fail. However, I do see the issue with pointer records.

jbcraig:
Well, my original thought didn't work either as the title includes the zone part.  The other option I see would be if the zone is set, but not the label then strip the zone from the title and use the remainder as the label.  This would allow simple and intuitive override of zone while still putting the fqdn in as the title.  Barring that, then how you have it seems to be a reasonable approach.

jyaworski:
What's the status of this PR? Can you rebase?

ekohl:
Closing due to inactivity.

PR #35, opened 2015-05-20
Use puppetlabs-concat instead of theforeman-concat_native

domcleal:


ekohl:
Untested, but :+1: code wise

mmoll:
not using -dns,  but looking good visually and tests are green :)
thanks @domcleal!

PR #38, opened 2015-07-30
support FreeBSD

mmoll:


templates/options.conf.erb
```diff
@@ -4,3 +4,6 @@ forwarders { <%= scope.lookupvar('::dns::forwarders').join("; ") %>; };
 <% end -%>
 
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
+<% if (@operatingsystem == 'FreeBSD') -%>
```
ekohl: Not sure if I like this. Is this needed?

mmoll: Yes, this is ugly. The FreeBSD package doesn't bring any RFC1918 zone files.

ekohl: But this is the pid file. I'd expect that it would be compiled with the sane default.

mmoll: argh, sorry... it's too early. The package is coming with a config file that's including this line so it's working out of the box. If this is a real problem, I can try to contact the package maintainer.

mmoll: He doesn't want to change the package for various reasons, so this would need to stay in (or handled via puppet and set if some variable is not undef)... your opinion?

ekohl: Let's leave it in for now then.

ekohl:
I think it looks good, but will have a closer look this weekend. Do remind me if I forget :)

ekohl:
Thanks!

PR #41, opened 2015-08-05
allow user to control allow-recursion on the global level

uriba107:
Allows a user to enable recursion on bind in the global level. useful for situation where multiple networks require access to a single server used as a full DNS server (not just for local zones).

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -12,9 +12,13 @@ class dns(
   $localzonepath        = $::dns::params::localzonepath,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $allow_recursion      = $::dns::params::allow_recursion,
   $namedconf_template   = $::dns::params::namedconf_template,
   $optionsconf_template = $::dns::params::optionsconf_template,
 ) inherits dns::params {
+  validate_array($dns::forwarders)
+  validate_array($dns::allow_recursion)
+
   class { '::dns::install': } ~>
   class { '::dns::config': } ~>
   class { '::dns::service': } ->
--- a/templates/options.conf.erb
+++ b/templates/options.conf.erb
@@ -4,6 +4,11 @@ forwarders { <%= scope.lookupvar('::dns::forwarders').join("; ") %>; };
 <% end -%>
 
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
+
+<% unless scope.lookupvar('::dns::allow_recursion').empty? -%>
+allow-recursion { <%= scope.lookupvar('::dns::allow_recursion').join("; ") %>; };
+<% end -%>
+
 <% if (@osfamily =~ /^(FreeBSD|DragonFly)$/) -%>
 pid-file "/var/run/named/pid";
 <% end -%>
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -12,8 +12,10 @@
   $localzonepath        = $::dns::params::localzonepath,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $allow_recursion      = $::dns::params::allow_recursion,
   $namedconf_template   = $::dns::params::namedconf_template,
   $optionsconf_template = $::dns::params::optionsconf_template,
+
```
ekohl: Nitpick: please remove this line

templates/options.conf.erb
```diff
@@ -4,6 +4,11 @@ forwarders { <%= scope.lookupvar('::dns::forwarders').join("; ") %>; };
 <% end -%>
 
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
+
+<% unless scope.lookupvar('::dns::allow_recursion').empty? -%>
```
ekohl: This doesn't allow to fully disable recursion, which IMHO is a missing feature right now. IMHO it would require a major version bump, but I'd be in favour of disabling recursion by default. The internet has become a hostile place and open recursors are a big problem due to reflection attacks.

manifests/init.pp
```diff
@@ -12,8 +12,10 @@
   $localzonepath        = $::dns::params::localzonepath,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $allow_recursion      = $::dns::params::allow_recursion,
```
ekohl: This can be validated with ```validate_array```.

ekohl:
I like this change. Inline comments to push it a bit further.

uriba107:
Any changes I need to do?
I've pretty much stayed in-form with what is used by the "forwarders" option.

mmoll:
@uriba107 see the inline comments for the changes we would like to see on top, then squash these changes into your existing commit and force push into your master branch, so this PR gets updated.

uriba107:
I've deleted the un-needed line. 
and added validate_array for both forwarders and allow_recursion in config.pp

manifests/config.pp
```diff
@@ -1,5 +1,8 @@
 # Configure dns
 class dns::config {
+  validate_array($dns::forwarders)
```
ekohl: In general we place these in init.pp. I've noticed this module is lacking them at all. Would you mind moving them?

templates/options.conf.erb
ekohl: I've noticed that [since Bind 9.4.1-P1 the default is safer](https://kb.isc.org/article/AA-00269/0/What-has-changed-in-the-behavior-of-allow-recursion-and-allow-query-cache.html) so now I think the current template is good.

ekohl:
:+1: even if you don't move the validate_array.

manifests/config.pp
uriba107: I opted to put them in config.pp as IMHO, it looked cleaner.
However, I have no problem moving them, your call.

ekohl: In all our modules we have it in init.pp so it would be more consistent. It's also closer to where the users passes it in.

manifests/init.pp
uriba107: Moved validate_array to init.pp as requested.

ekohl:
Thanks!

PR #42, opened 2015-08-11
Add options and zone possibilities

AblionGE:
I added diverse options for bind that can be necessary to manage your own DNS :
- recursion
- allow_query
- dnssec_enable
- dnssec_validation

And also the possibility to not create a zone file when you declare a zone (only add an entry in the conf file)

An finally, I adapted the templates.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -12,12 +12,20 @@ class dns(
   $localzonepath        = $::dns::params::localzonepath,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $recursion            = $::dns::params::recursion,
   $allow_recursion      = $::dns::params::allow_recursion,
+  $allow_query          = $::dns::params::allow_query,
+  $dnssec_enable        = $::dns::params::dnssec_enable,
+  $dnssec_validation    = $::dns::params::dnssec_validation,
   $namedconf_template   = $::dns::params::namedconf_template,
   $optionsconf_template = $::dns::params::optionsconf_template,
 ) inherits dns::params {
   validate_array($dns::forwarders)
   validate_array($dns::allow_recursion)
+  validate_array($dns::allow_query)
+  validate_re($dns::recursion, '^(yes|no)$', 'Only \'yes\' and \'no\' are valid values for recursion field')
+  validate_re($dns::dnssec_enable, '^(yes|no)$', 'Only \'yes\' and \'no\' are valid values for dnssec_enable field')
+  validate_re($dns::dnssec_validation, '^(yes|no|auto)$', 'Only \'yes\', \'no\' and \'auto\' are valid values for dnssec_validation field')
 
   class { '::dns::install': } ~>
   class { '::dns::config': } ~>
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -57,5 +57,10 @@ class dns::params {
 
     $listen_on_v6         = 'any'
 
+    $recursion            = 'yes'
     $allow_recursion      = []
+    $allow_query          = [ 'any' ]
+
+    $dnssec_enable        = 'yes'
+    $dnssec_validation    = 'yes'
 }
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -16,10 +16,14 @@ define dns::zone (
     $contact        = "root.${title}.",
     $zonefilepath   = $::dns::zonefilepath,
     $filename       = "db.${title}",
+    $manage_file    = true,
+    $forward        = 'first',
+    $forwarders     = [],
 ) {
 
-  validate_bool($reverse)
-  validate_array($masters, $allow_transfer)
+  validate_bool($reverse, $manage_file)
+  validate_array($masters, $allow_transfer, $forwarders)
+  validate_re($forward, '^(first|only)$', 'Only \'first\' or \'only\' are valid values for forward field')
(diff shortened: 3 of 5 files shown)


ekohl:
Looks like it needs a rebase.

manifests/init.pp
```diff
@@ -12,9 +12,13 @@
...
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $recursion            = $::dns::params::recursion,
   $allow_recursion      = $::dns::params::allow_recursion,
+  $allow_query          = $::dns::params::allow_query,
+  $dnssec_enable        = $::dns::params::dnssec_enable,
+  $dnssec_validation    = $::dns::params::dnssec_validation,
   $namedconf_template   = $::dns::params::namedconf_template,
-  $optionsconf_template = $::dns::params::optionsconf_template,
+  $optionsconf_template = $::dns::params::optionsconf_template
```
mmoll: the comma can stay, as it eases future extensions ;)

mmoll:
I didn't test this, but it's looking correct code-wise, just one nitpick.
Maybe it would make sense to use booleans for the yes/no values?

AblionGE:
For the yes/no values, my idea was to be close of the actual values in the bind documentation.

For example : http://www.zytrax.com/books/dns/ch7/queries.html#recursion

I wanted to write tests, but I don't know when I will have time fot that...

mmoll:
For the yes/no values, my idea is to use the boolean type in puppet because of validation and the whole handling regarding type safety (if one may call it like that) in Puppet.

Regarding tests, this module has an embarrassingly low coverage, so I guess we can merge this without spec tests for now and cry later :cry:.

mmoll:
Oh, and please squash your commits together into one.

AblionGE:
Squash done!
About the validation of yes/no values, I added some validations of regular expressions in the init class to enforce it. This remains a string but I think it's better to stay close of the actual values of bind.

manifests/init.pp
```diff
@@ -12,12 +12,20 @@
...
+  $allow_query          = $::dns::params::allow_query,
+  $dnssec_enable        = $::dns::params::dnssec_enable,
+  $dnssec_validation    = $::dns::params::dnssec_validation,
   $namedconf_template   = $::dns::params::namedconf_template,
   $optionsconf_template = $::dns::params::optionsconf_template,
 ) inherits dns::params {
   validate_array($dns::forwarders)
   validate_array($dns::allow_recursion)
+  validate_array($dns::allow_query)
+  validate_re($dns::recursion, ['yes', 'no'], 'Only \'yes\' and \'no\' are valide values for recursion field')
```
domcleal: a) use ^ and $ in regexes to anchor it, e.g. `'^yes$'` or even combine them with '^(yes|no)$'.
b) "valide" should be "valid"

domcleal:
I don't really have an opinion on bool/string I'm afraid.  I would usually prefer boolean, but it wouldn't surprise me if BIND added a third value to those fields in future, so string allows for a bit more flexibility.

We don't seem to have any precedent in this module that I can see to help..

manifests/init.pp
AblionGE: Yes of course ;)

AblionGE:
Exactly, that was what I had in mind. Future extension of BIND will be easier to include

manifests/zone.pp
```diff
@@ -16,10 +16,14 @@
...
+    $file           = true,
+    $forward        = 'first',
+    $forwarders     = [],
 ) {
 
-  validate_bool($reverse)
-  validate_array($masters, $allow_transfer)
+  validate_bool($reverse, $file)
+  validate_array($masters, $allow_transfer, $forwarders)
+  validate_re($forward, ['first', 'only'], 'Only \'first\' or \'only\' are valid values for forward field')
```
mmoll: '^(first|only)$' then here also, please

mmoll:
one last validate_re, then I'm :+1:

AblionGE:
Done :+1:

manifests/zone.pp
```diff
@@ -16,10 +16,14 @@
     $contact        = "root.${title}.",
     $zonefilepath   = $::dns::zonefilepath,
     $filename       = "db.${title}",
+    $file           = true,
```
ekohl: This is somewhat generic. Maybe name it ```manage_file``` or something similar?

manifests/params.pp
```diff
@@ -57,5 +57,10 @@
 
     $listen_on_v6         = 'any'
 
+    $recursion            = 'yes'
     $allow_recursion      = []
+    $allow_query          = [ 'any' ]
+
+    $dnssec_enable        = 'no'
+    $dnssec_validation    = 'no'
```
ekohl: http://ftp.isc.org/isc/bind9/cur/9.8/doc/arm/Bv9ARM.ch06.html states the default is ```yes``` and I'd prefer to match defaults.

ekohl:
For the record, I agree with yes/no instead of booleans.

manifests/params.pp
AblionGE: It seems obvious ;)

manifests/zone.pp
AblionGE: :+1:

ekohl:
:+1:

mmoll:
merged, merci @AblionGE!

PR #46, opened 2015-09-21
use ensure_packages

mmoll:
I'm wondering if the test shouldn't be something like `should contain_ensure_packages(['bind'])`, but that doesn't work.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,7 +1,4 @@
 # Install dns service
 class dns::install {
-  package { 'dns':
-    ensure => present,
-    name   => $dns::dns_server_package,
-  }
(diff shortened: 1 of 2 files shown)


domcleal:
No, I think this is exactly correct.  Because ensure_packages is a parser function, it's evaluated during catalog compilation so the catalog that rspec-puppet is testing simply contains a package resource.  There's no evidence in the catalog that it used ensure_packages.

mmoll:
perfect, my job is done here ;)

domcleal:
Merged, thanks @mmoll.  I upped the dep on stdlib to 2.6 which introduced the function.

PR #48, opened 2015-10-28
Allow control of the running service

sd-robbruce:


ekohl:
I think you also want to control the ```enable``` parameter because otherwise it still starts at boot.

mmoll:
:+1:

ekohl:
Thanks!

PR #49, opened 2015-11-10
Do not set DNS as an insecure open recurive server by default

kmcfate:
The default settings create an open DNS server which allows a very significant attack vector for DNS spoofing attacks.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -57,7 +57,7 @@ class dns::params {
 
     $listen_on_v6         = 'any'
 
-    $recursion            = 'yes'
+    $recursion            = 'no'
     $allow_recursion      = []
     $allow_query          = [ 'any' ]
 


ekohl:
While I agree this could be a good change, I believe this is mitigated by the fact bind by default only allows recursion from local networks. In fact, we rely on this in our default setup dhcp enabled setup: https://github.com/theforeman/puppet-foreman_proxy/blob/bfc2c8a514c51802d0f704ee1bbdcc3b3ff09f3d/manifests/proxydhcp.pp#L21-L25. That makes me hesitant to just change it.

kmcfate:
I can confirm that bind-9.8.2-0.37.rc1.el6_7.4.x86_64 allows all recursive requests by default.

ekohl:
https://kb.isc.org/article/AA-00269/0/What-has-changed-in-the-behavior-of-allow-recursion-and-allow-query-cache.html says that prior to Bind 9.4.1-P1 the default policy was allow all, but it's been changed to ```localnets; localhost```. I'd expect that your version would have the newer defaults so a bug report upstream might be needed.

ekohl:
@kmcfate any progress on this?

mmoll:
no reaction -> closing.

ekohl:
I can confirm this is a problem. The problem is that we explicitly set ```allow_query``` to any. This in turn makes it an open resolver which is a very bad thing for the internet. I'll submit a PR to match saner defaults bind has nowadays.

PR #52, opened 2015-12-11
Add the forward [ only | first ] option at the global level

jeyglk:
There is already the ability to configure forwarders, but the forward option is missing - therefore adding it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -10,6 +10,7 @@ class dns(
   $namedservicename     = $::dns::params::namedservicename,
   $zonefilepath         = $::dns::params::zonefilepath,
   $localzonepath        = $::dns::params::localzonepath,
+  $forward              = $::dns::params::forward,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
   $recursion            = $::dns::params::recursion,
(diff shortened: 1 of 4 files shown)


mmoll:
@jeyglk could you add a positive and negative spec test, please? :mag:

mmoll:
Oh, and plase push changes into your master branch in this case... there's no need to re-fork and open new a pull request after each change.

Normally, after forking a repo, one does use a topic-branch for such changes and open the PR from that branch. This way it's possible to re-sync with the current master branch from the upstream fork. If you have any questions about the workflow, there's a lot of documentation on help.github.com, but also don't hesitate to ask here!

jeyglk:
@mmoll updated

mmoll:
Maybe adding a validate_re for '(only|first)' would be nice if the variable is set, :+1: otherwise.

jeyglk:
@mmoll Updated again.

manifests/init.pp
```diff
@@ -30,6 +31,9 @@
   validate_re($dns::dnssec_enable, '^(yes|no)$', 'Only \'yes\' and \'no\' are valid values for dnssec_enable field')
   validate_re($dns::dnssec_validation, '^(yes|no|auto)$', 'Only \'yes\', \'no\' and \'auto\' are valid values for dnssec_validation field')
   validate_re($dns::service_ensure, '^running|true|stopped|false$', 'Only \'running\', \'false\', \'stopped\' and \'false\' are validate values for service_ensure field')
+  if $dns::forward {
```
ekohl: You can just use $forward here (and the line below).

ekohl:
:+1: with the minor tweak. Would you mind squashing your commits?

manifests/init.pp
jeyglk: I don't mind modifying it, but the prefix is written everywhere else, I did it to match the other lines.

ekohl: You're right, consistency is better.

mmoll:
merged, merci @jeyglk!

PR #53, opened 2015-12-24
Limit recursion to localnets and localhost

ekohl:
This doesn't turn the server into an open recursor by default and
matches the defaults introduced in BIND 9.4.1-P1.

https://kb.isc.org/article/AA-00269/0/What-has-changed-in-the-behavior-of-allow-recursion-and-allow-query-cache.html

Reported in https://github.com/theforeman/puppet-dns/pull/49 by @kmcfate

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -58,7 +58,7 @@ class dns::params {
     $listen_on_v6         = 'any'
 
     $recursion            = 'yes'
-    $allow_recursion      = []
+    $allow_recursion      = [ 'localnets', 'localhost' ]
     $allow_query          = [ 'any' ]
 
     $empty_zones_enable   = 'yes'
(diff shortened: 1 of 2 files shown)


ekohl:
Verified on EL6 with bind-9.8.2-0.37.rc1.el6_7.5.x86_64

mmoll:
:+1: maybe adding a test that checks the content of the generated file would be good?

ekohl:
Updated

mmoll:
merged, bedankt @ekohl!
@kmcfate it would be nice if you could test this change and report back.

ekohl:
@domcleal I think we should consider this a security issue. Like @kmcfate correctly states, open recursors are a major threat and can be leveraged to perform DDoS attacks.

PR #54, opened 2016-01-20
Add dns_notify to dns class and dns::zone define

treydock:
In a situation with a single master and multiple slaves, the slaves will end up attempting to notify other slaves of updates to the zones.  Messages like this will appear

```
Jan 20 10:53:33 HOSTNAME named[1628]: client SLAVE_IP#52691: received notify for zone 'DOMAIN'
Jan 20 10:53:33 HOSTNAME named[1628]: zone DOMAIN/IN: refused notify from non-master: SLAVE_IP#52691
```

It may be worth making the value for `notify` automatically `no` when `zonetype` is `slave`, but wasn't sure if such an assumption would be too much.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -20,13 +20,23 @@ define dns::zone (
     $manage_file    = true,
     $forward        = 'first',
     $forwarders     = [],
+    $dns_notify     = undef,
 ) {
 
   validate_bool($reverse, $manage_file)
   validate_array($masters, $allow_transfer, $forwarders, $also_notify)
   validate_re($forward, '^(first|only)$', 'Only \'first\' or \'only\' are valid values for forward field')
+  if $dns_notify {
+    validate_re($dns_notify, '^(yes|no|explicit)$', 'Only \'yes\', \'no\', or \'explicit\' are valid values for dns_notify field')
+  }
 
   $zonefilename = "${zonefilepath}/${filename}"
+
+  if $zonetype == 'slave' {
+    $_dns_notify = pick($dns_notify, 'no')
+  } else {
+    $_dns_notify = $dns_notify
+  }
 
   concat::fragment { "dns_zones+10_${zone}.dns":
     target  => $::dns::publicviewpath,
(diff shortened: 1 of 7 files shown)


domcleal:
:+1: looks OK to me, and I'd be fine with changing the default to "no" for slaves.

manifests/zone.pp
```diff
@@ -20,6 +20,7 @@
     $manage_file    = true,
     $forward        = 'first',
     $forwarders     = [],
+    $dns_notify     = undef,
```
ekohl: You could derive this in the code, if type is slave you can set it to no.

treydock:
I've updated PR to set notify to "no" for slaves, if the dns_notify parameter is left undefined.

domcleal:
Looks good :+1:

templates/named.zone.erb
```diff
@@ -23,4 +23,7 @@ zone "<%= @zone %>" {
 <% unless @forwarders.empty? -%>
     forwarders { <%= @forwarders.join('; ') %>; };
 <% end -%>
+<% if @_dns_notify -%>
```
ekohl: Will this work with all undefined/empty values?

manifests/zone.pp
```diff
@@ -28,6 +29,12 @@
 
   $zonefilename = "${zonefilepath}/${filename}"
 
+  if $zonetype == 'slave' {
+    $_dns_notify = pick($dns_notify, 'no')
```
ekohl: I recall at some point puppet doesn't like ```$_``` variables, but I'm not 100% sure about it.

templates/named.zone.erb
domcleal: This should work well for undef, but it could write an empty string if set to `''` I think.  Perhaps a `validate_re` in the manifest would catch that.

treydock:
@domcleal I've updated PR to use validate_re in both `dns` class and `dns::zone` defined type.

@ekohl I have been using `$_` on numerous modules to denote "private" variables.  So far it has not caused me problems in 3.6.x or 3.8.x.  I run many of those modules through future parser and Puppet 4 unit tests and have seen no issues.  If the use of `$_` is a problem for Puppet, I'd very much like to know as I'd have to update most of my modules, as well as this PR.

ekohl:
@treydock like I said, it was a vague recollection. In that case I trust you on it since it's trivial to change if it does turn out to be a problem somewhere.

:+1:

mmoll:
merged, thanks @treydock!

PR #55, opened 2016-02-17
Update min Puppet version and supported OSes

domcleal:


mmoll:
cherry-picked as f9bd385 with Ubuntu/xenial added, thanks @domcleal!

PR #56, opened 2016-02-24
Add ACL functionality to named.conf template

jyaworski:


templates/named.conf.erb
```diff
@@ -14,5 +14,15 @@ options  {
 include "<%= scope.lookupvar('::dns::localzonepath') %>";
 <% end -%>
 
+<%- if scope.lookupvar('::dns::acl_hash') and !scope.lookupvar('::dns::acl_hash').empty? -%>
```
ekohl: Isn't this redundant? We know it's a hash since validate_hash checked that for us. If it's empty, the loop iterator will just do nothing.

ekohl:
It makes sense, but we're really lacking in documentation and this parameter needs some. I don't it's obvious that it needs to be a hash containing string -> list.

templates/named.conf.erb
jyaworski: Keeping the check for .empty? ensures we don't have the acl block in the config when the hash is empty.

jyaworski:
Docs added.

ekohl:
:+1: for the code. Would you mind adding a test to dns_init_spec.rb?

jyaworski:
@ekohl will do so once #57 is merged and I can rebase.

mmoll:
please rebase :)

jyaworski:
@mmoll rebased.

manifests/init.pp
```diff
@@ -1,5 +1,30 @@
...
+#                          and 'allowed_addresses' => array
+#                          type:hash
+#
+# === Usage:
+#
+# * Simple usage:
+#
+#     include dns
+#
+# * Advanced usage:
```
ekohl: I think you intended to add an example here.

templates/named.conf.erb
```diff
@@ -16,5 +16,15 @@ options  {
 include "<%= scope.lookupvar('::dns::localzonepath') %>";
 <% end -%>
 
+<%- unless scope.lookupvar('::dns::acls').empty? -%>
```
ekohl: I still think this line is redundant: if it's empty you get 0 iterations of the loop and no output, right?

jyaworski: The reason for this line is to have no acl block at all if the array is empty as to not clutter the config.

ekohl: Like I said, this line is redundant. If acls is empty, the loop won't iterate and no block is shown. You can remove it and the 'with no custom parameters' test still passes. It contains no acl block.

jyaworski: Removed.

manifests/init.pp
ekohl: If you include the example here (or remove the Advanced usage:) it's good to go.

jyaworski: Removed.

jyaworski:
I believe I've addressed all of the issues raised, and this is in a merge-able state.

ekohl:
Thanks!

PR #59, opened 2016-04-15
Change default path for zones.conf to /etc/named/zones.conf for RedHat

treydock:
Set default path for zones.conf to /etc/named/zones.conf to allow easier use of named with chroot enabled

I don't like the fact that existing systems will have /etc/zones.conf left behind unless a site was passing the `publicviewpath` parameter.  Wasn't sure how to best handle changing default path.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -7,6 +7,7 @@ class dns::params {
         $optionspath        = "${dnsdir}/named.conf.options"
         $zonefilepath       = "${vardir}/zones"
         $localzonepath      = "${dnsdir}/zones.rfc1918"
+        $publicviewpath     = "${dnsdir}/zones.conf"
         $dns_server_package = 'bind9'
         $namedservicename   = 'bind9'
         $user               = 'bind'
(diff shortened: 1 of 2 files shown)


ekohl:
I think this was based on the default RH config where it uses ```/etc/named.rfc1912.zones``` and ```/etc/named.root.key```, but I agree that's a bad default.

This is a major version bump IMHO, but since puppet manages this file I think that's sufficient.

domcleal:
:+1: 

> This is a major version bump IMHO, but since puppet manages this file I think that's sufficient.

You mean a major bump is sufficient, or something else?  I'd say a minor is probably OK.

ekohl:
>> This is a major version bump IMHO, but since puppet manages this file I think that's sufficient.
> 
> You mean a major bump is sufficient, or something else? I'd say a minor is probably OK.

I thought it was a big change for admins so a major bump, but since puppet manages the file that would be enough. Now that I think about it you may be right that a minor is sufficient. The "API" doesn't change.

mmoll:
:+1: for merging and after that doing a minor bump.

ekohl:
Thanks!

PR #61, opened 2016-06-09
ensure_packages still results in dup definition

andybrucenet:
I have a class that wants to use the puppet-dns class. In *my* class, I have:
`ensure_packages(['bind', 'bind-utils'], {ensure => 'installed'})`
There is also an 'ensure_packages' test inside install.pp; the two ensure_packages calls should have worked together to avoid duplicate package definition errors.
However, I received:
`Error: Duplicate declaration: Package[bind] is already declared; cannot redeclare at /root/proj/puppet/modules/dns/manifests/install.pp:3`
I first tried to set the variable name to empty:
`    class {
      '::dns':
        dns_server_package => '';
    }
`
However, that simply caused a puppet syntax error. So I made the change to test for an empty package name and avoid ensure_packages altogether in this fork.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -1,4 +1,6 @@
 # Install dns service
 class dns::install {
-  ensure_packages([$dns::dns_server_package])
+  if ! empty($dns::dns_server_package) {
+    ensure_packages([$dns::dns_server_package])
+  }
 }


domcleal:
The reason for the duplicate is the `ensure => installed`, if you used `present` or left it (present's the default) then they wouldn't conflict. The ensure_packages function checks for these differences in params/properties.

I'm OK with this change though, so :+1:.  I would ask that you add a summary line to the start of the commit message so it doesn't start with "In my tests [..]", e.g. `Skip package installation when dns_server_package is empty` by amending it (`git commit --amend`).

andybrucenet:
Done - thank you for the education!

domcleal:
Merged as b407cb9, thanks @andybrucenet!

PR #63, opened 2016-06-16
Nest 'forwarders' only if 'forward' is used.

jkrauska:
forwarders is only relevant if the zone is forwarded zone.  

This change only writes out 'forwarders' for forward domains.


domcleal:
:+1: appears correct according to the 9.4 documentation.

ekohl:
Thanks!

PR #64, opened 2016-07-05
Use concat fragments instead of file resource templates for config

dmsimard:
The net effect of this change is zero - it's backwards compatible
and does not result in any change for users.
The concat resources create the files with the same permissions
and the concat fragments take care of inserting the templates with
the same feature to override the template if need be.

However, since concat is now used, consumers of puppet-dns can now
insert their own concat fragments without having to maintain their
own template.
Before this change, if we wanted to diverge slightly from the
template, we would need to "fork" the template and maintain our own.
We could otherwise edit the file with resources like file_line but
this is not idempotent.

Spec test coverage for these bits have been improved a bit and
standardized as part of the commit.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -7,28 +7,36 @@ class dns::config {
     group => $dns::params::group,
     mode  => '0640',
   }
+
   concat::fragment { 'dns_zones+01-header.dns':
     target  => $::dns::publicviewpath,
     content => ' ',
     order   => '01',
   }
 
-  file {
-    $dns::namedconf_path:
-      owner   => root,
-      group   => $dns::params::group,
-      mode    => '0640',
-      content => template($::dns::namedconf_template);
-    $dns::optionspath:
-      owner   => root,
-      group   => $dns::params::group,
-      mode    => '0640',
-      content => template($::dns::optionsconf_template);
-    $dns::zonefilepath:
-      ensure  => directory,
-      owner   => $dns::params::user,
-      group   => $dns::params::group,
-      mode    => '0640';
+  concat { [$::dns::namedconf_path, $::dns::optionspath]:
+    owner => root,
+    group => $::dns::params::group,
+    mode  => '0640',
+  }
+
+  concat::fragment { 'named.conf+10-main.dns':
+    target  => $::dns::namedconf_path,
+    content => template($::dns::namedconf_template),
+    order   => '10',
+  }
+
+  concat::fragment { 'options.conf+10-main.dns':
+    target  => $::dns::optionspath,
+    content => template($::dns::optionsconf_template),
+    order   => '10',
+  }
+
+  file { $dns::zonefilepath:
+    ensure => directory,
+    owner  => $dns::params::user,
+    group  => $dns::params::group,
+    mode   => '0640',
   }
 
   exec { 'create-rndc.key':
@@ -41,4 +49,3 @@ class dns::config {
     mode  => '0640',
   }
 }
-
--- a/spec/classes/dns_init_spec.rb
+++ b/spec/classes/dns_init_spec.rb
@@ -20,13 +20,8 @@ describe 'dns' do
 
       it { should contain_package('bind').with_ensure('present') }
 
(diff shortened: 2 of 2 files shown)


dmsimard:
Results for the two templates..

Before:

    Notice: /Stage[main]/Dns::Config/File[/etc/named/options.conf]/content: content changed '{md5}0f8469748b8a0403c3d687976370f180' to '{md5}d10d4b5c17906eb720d9965d80d1e98f'
    Notice: /Stage[main]/Dns::Config/File[/etc/named.conf]/content: content changed '{md5}c4b43a63125a3d365700e73ad5161b96' to '{md5}0a81a69c667dc73ac1b4133b5a732575'
    Notice: /Stage[main]/Dns::Service/Service[named]: Triggered 'refresh' from 1 events

After:

    Notice: /Stage[main]/Dns::Config/Concat[/etc/named.conf]/File[/etc/named.conf]/content: content changed '{md5}c4b43a63125a3d365700e73ad5161b96' to '{md5}0a81a69c667dc73ac1b4133b5a732575'
    Notice: /Stage[main]/Dns::Config/Concat[/etc/named/options.conf]/File[/etc/named/options.conf]/content: content changed '{md5}0f8469748b8a0403c3d687976370f180' to '{md5}d10d4b5c17906eb720d9965d80d1e98f'
    Notice: /Stage[main]/Dns::Service/Service[named]: Triggered 'refresh' from 1 events

manifests/config.pp
```diff
@@ -13,22 +13,29 @@
...
-      owner   => root,
-      group   => $dns::params::group,
-      mode    => '0640',
-      content => template($::dns::optionsconf_template);
-    $dns::zonefilepath:
-      ensure  => directory,
-      owner   => $dns::params::user,
-      group   => $dns::params::group,
-      mode    => '0640';
+  concat { [$::dns::namedconf_path, $::dns::optionspath]:
```
EmilienM: skip a line at LINE 10

EmilienM:
:+1: excellent work here.

ekohl:
I would be open to such a change. I'll let Travis do it's thing, but I think :+1: once it's green.

iurygregory:
Awesome work David =D

dmsimard:
@ekohl thanks! Looks like travis is green :)

manifests/config.pp
```diff
@@ -7,28 +7,36 @@
...
+    owner => root,
+    group => $::dns::params::group,
+    mode  => '0640',
+  }
+
+  concat::fragment { 'namedconf_template':
```
domcleal: Use the existing naming schema for fragments, i.e. `named.conf+10-main.dns` or similar.

manifests/config.pp
```diff
@@ -7,28 +7,36 @@
...
+    target  => $::dns::namedconf_path,
+    content => template($::dns::namedconf_template),
+    order   => '10',
+  }
+
+  concat::fragment { 'optionsconf_template':
```
domcleal: ditto, `options.conf+10-main.dns`

manifests/config.pp
dmsimard: fixed

manifests/config.pp
dmsimard: fixed

domcleal:
:+1: looks fine now, thanks.

dmsimard:
Missed something with my find/replace in the spec tests, should be good now.

dmsimard:
@domcleal @ekohl I think we're good now, thanks for your help. Would love to see this in ASAP, it'd unblock an ongoing effort on our end.

brandonweeks:
@dmsimard merged, thanks!

EmilienM:
@brandonweeks do you have any ETA before next release?

mmoll:
IMHO we can cut a 3.3.1 from current master any time... anybody from @theforeman: any objections?

ekohl:
No objection from my side

mmoll:
Released 3.3.1 to the forge.

Issue #67, opened 2016-09-27
It's not possible to have BIND not listen on port 53 in ipv6


dmsimard:
If I want to have BIND listen on a different port on v4, I can do the following:

    class { '::dns':
      additional_options => {
        'listen-on' => 'port 5322 { any; }',
      }
    }

However, for ipv6, there's the listen_on_v6 parameter that will **always** be present in the [template](https://github.com/theforeman/puppet-dns/blob/c4e4afd431565c28bb892f0261928baaa63d2dfa/templates/options.conf.erb#L19):

    listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };

There's no way to precise the port so it will always be on port 53.

For backwards compatibility purposes, I recommend adding a condition in the template -- if listen_on_v6 is defined, add it with the value.
If someone wants more "advanced" config, such as listening on v6 on an alternate port, he can set listen_on_v6 to false and use additional_options like the snippet at the beginning of this issue:

    class { '::dns':
      listen_on_v6       => false,
      additional_options => {
        'listen-on'    => 'port 5322 { any; }',
        'listen-on-v6' => 'port 5322 { any; }'
      }
    }

PR #68, opened 2016-09-27
Do not configure listen-on-v6 parameter if it's not set

dmsimard:
The listen-on-v6 parameter would always be configured in
options.conf. This makes it unflexible with the use case of making
BIND listen on another port than 53.
Make it so the configuration is added only when the parameter is set
so that, if required, a user can set listen_on_v6 to false and add
his own configuration in additional_options.

Closes: #67

### DIFF
--- a/templates/options.conf.erb
+++ b/templates/options.conf.erb
@@ -16,7 +16,9 @@ empty-zones-enable <%= scope.lookupvar('::dns::empty_zones_enable') %>;
 <% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::dns_notify')) -%>
 notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% end -%>
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
+<% end -%>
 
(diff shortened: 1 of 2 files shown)


dmsimard:
@theforeman: Would love to see this rolled into a dot release ASAP, need it for an implementation in OpenStack Continuous Integration with DNS as a Service. Thanks :)

templates/options.conf.erb
```diff
@@ -16,7 +16,9 @@ empty-zones-enable <%= scope.lookupvar('::dns::empty_zones_enable') %>;
 <% unless [nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::dns_notify')) -%>
 notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% end -%>
+<% if scope.lookupvar('::dns::listen_on_v6') -%>
```
ekohl: I think it'd be best if we also checked if it's non-empty.

ekohl:
In general :+1:, just a minor nitpick.

dmsimard:
Thanks @ekohl, updated with your feedback.

domcleal:
Thanks @dmsimard.

dmsimard:
@domcleal awesome! Do you believe we could get a new version tag with that in? Would appreciate very much.

domcleal:
3.4.1's tagged and release to Puppet Forge.

Issue #66, opened 2016-09-23
Minor glitch in README


hguemar:
README says that this project is licensed under GPLv3+ but LICENSE and metadata says ASL 2.0, could you fix it, please?

mmoll:
@GregSutcliffe as you did the inital fork and import, could you have a look?

GregSutcliffe:
The original module was from Zach Leslie, IIRC, but it seems to be gone from his GitHub repos now. Assuming that's now https://github.com/xaque208/puppetlabs-dns then the original appears to be Apache 2.0 - but that module doesn't support Bind, so maybe that's not it.

So, perhaps the best thing to do is ask :)

@xaque208 This is going back to 2012 but do you happen to recall the license for your old DNS module? Or, to ask a different question, do you have any concerns about clarifying the license on this derivative module?

GregSutcliffe:
@xaque208 scratch that, the history has been traced - ty anyway!

xaque208:
Oh gad!  I don't think the bind module ever made it public, though it should have.  Glad you sorted it out, as that was ages ago.  All licensing for PuppetLabs items should be Apache2.  Cheers @GregSutcliffe.

PR #69, opened 2016-10-04
Change licence text in README to match LICENSE/metadata

domcleal:
The commit history shows the module has always been under the ASL-2.0,
and that the README text was an inconsistency in 999eb91 where the
LICENSE document was added and README.md was extended (probably from a
template).

Updated the main LICENSE document to include the full text, not only the
boilerplate copyright statement.

Fixes GH-66

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,11 +1,202 @@
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
 
-http://www.apache.org/licenses/LICENSE-2.0
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


LICENSE
```diff
@@ -1,11 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
ekohl: Should this be filled in?

domcleal: Well this is part of the appendix showing how to use the licence rather than being _the_ copyright notice, so no, but it _is_ in README.md. Often people just leave the appendices in their licence files rather than removing it.

ekohl: That makes sense. Now you can also exactly match the text rather than a fuzzy match.

domcleal: Yes, I think the licensee gem (used by GitHub for the licence identifier in the top-right of a repo) uses an exact match on the file contents.

ekohl:
Thanks!

PR #72, opened 2016-12-06
Added additional_files and additional_config options

msweetser:
I added the options additional_files and additional_config to fill a need of mine using the module.

additional_files will be added as "include" directives, one per file.

additional_config will be added as key value pairs.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -49,6 +49,7 @@ class dns (
   $service_ensure       = $::dns::params::service_ensure,
   $service_enable       = $::dns::params::service_enable,
   $additional_options   = $::dns::params::additional_options,
+  $additional_files     = $::dns::params::additional_files,
 ) inherits dns::params {
   validate_array($dns::forwarders)
   validate_array($dns::allow_recursion)
@@ -67,6 +68,7 @@ class dns (
   validate_hash($controls)
   validate_hash($acls)
   validate_hash($additional_options)
+  validate_array($additional_files)
 
   class { '::dns::install': } ~>
   class { '::dns::config': } ~>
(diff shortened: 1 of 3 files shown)


ekohl:
How does ```additional_config``` differ from ```additional_options```? The only difference I see is that it's placed in a different file.

Please also fix the trailing whitespace that's causing the build to fail.

ekohl:
For ```additional_files```, have you considered using ```concat::fragment``` instead?

dlucredativ:
> How does additional_config differ from additional_options? The only difference I see is that it's placed in a different file.

I have a use case for additional_files as well. There are directives that need to go top-scope, like logging, additional keys and such. I was about to create a pull request myself, then I found this one. Can I help somehow? @ekohl, would you rather have the additional_file content included in the named.conf via fragments?

ekohl:
@dlucredativ I'm not against it per se, but I'd like to know the use case and know why the fragments solution would be inadequate.

dlucredativ:
@ekohl, my use case is to configure additional top-scope bind directives (logging, keys). I am fine with `include` statements like in this PR as well as with fragment-based injection into `named.conf`. If it helps the case, I can open another PR with fragments and a fixed travis build. I don't know if @msweetser insists on includes, though.

ekohl:
@dlucredativ if you're managing those additional directives through puppet then I'd suggest you use fragments.

dlucredativ:
@ekohl to be clear, are you suggesting to implement "additional_directives" with `concat::fragment` as a new puppet-dns feature, or to use `concat::fragment` outside of puppet-dns?

ekohl:
@dlucredativ I mostly meant that for uncommon uses to use ```concat::fragment``` outside of puppet-dns, but I think we'd accept a patch that adds ```additional_directives```. However, I think that can be implemented in the main ```named.conf``` with a simple for loop.

dlucredativ:
Alright, I will create a pull request with additional_directives in a few moments.

ekohl:
Closing in favour of #79

PR #74, opened 2016-12-09
Add some attribues config 

jas01:
I add some variables in the config file like the listen-on (for example when you server get multiple ip address and you don't want bind listen on all interface), same problem for notify_source and transfert_source. 

I think it's better not to put that inside a additionnel_options because it's 3 fields.


jas01:
Shit...sorry wrong version...mix up

PR #75, opened 2016-12-09
Add some variables

jas01:
I add some variables in the config file like the listen-on (for example when you server get multiple ip address and you don't want bind listen on all interface), same problem for notify_source and transfert_source. 

I think it's better not to put that inside a additionnel_options because it's 3 fields.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -35,7 +35,7 @@ class dns (
   $forward              = $::dns::params::forward,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
-  $listen_on_v4         = $::dns::params::listen_on_v4,
+  $listen_on            = $::dns::params::listen_on,
   $recursion            = $::dns::params::recursion,
   $allow_recursion      = $::dns::params::allow_recursion,
   $allow_query          = $::dns::params::allow_query,
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -35,6 +35,7 @@ class dns (
   $forward              = $::dns::params::forward,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
+  $listen_on_53         = $::dns::params::listen_on_53,
   $recursion            = $::dns::params::recursion,
   $allow_recursion      = $::dns::params::allow_recursion,
   $allow_query          = $::dns::params::allow_query,
@@ -49,6 +50,11 @@ class dns (
   $service_ensure       = $::dns::params::service_ensure,
   $service_enable       = $::dns::params::service_enable,
   $additional_options   = $::dns::params::additional_options,
+  $transfert_source     = $::dns::params::transfert_source,
+  $transfert_source_port= $::dns::params::transfert_source_port,
+  $notify_source        = $::dns::params::notify_source,
+  $notify_source_port   = $::dns::params::notify_source_port
+
 ) inherits dns::params {
   validate_array($dns::forwarders)
   validate_array($dns::allow_recursion)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -50,10 +50,10 @@ class dns (
   $service_ensure       = $::dns::params::service_ensure,
   $service_enable       = $::dns::params::service_enable,
   $additional_options   = $::dns::params::additional_options,
-  $transfert_source     = $::dns::params::transfert_source,
-  $transfert_source_port= $::dns::params::transfert_source_port,
+  $transfer_source      = $::dns::params::transfer_source,
+  $transfer_source_port = $::dns::params::transfer_source_port,
   $notify_source        = $::dns::params::notify_source,
-  $notify_source_port   = $::dns::params::notify_source_port
+  $notify_source_port   = $::dns::params::notify_source_port,
 
 ) inherits dns::params {
   validate_array($dns::forwarders)
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -35,7 +35,7 @@ class dns (
   $forward              = $::dns::params::forward,
   $forwarders           = $::dns::params::forwarders,
   $listen_on_v6         = $::dns::params::listen_on_v6,
-  $listen_on_53         = $::dns::params::listen_on_53,
+  $listen_on_v4         = $::dns::params::listen_on_v4,
   $recursion            = $::dns::params::recursion,
   $allow_recursion      = $::dns::params::allow_recursion,
   $allow_query          = $::dns::params::allow_query,
--- a/templates/named.zone.erb
+++ b/templates/named.zone.erb
@@ -6,9 +6,7 @@ zone "<%= @zone %>" {
     forwarders { <%= @forwarders.join('; ') %>; };
 <% end -%>
 <% end -%>
-<% if @manage_file == true -%>
     file "<%= @zonefilename %>";
-<% end -%>
 <% if @zonetype == 'master' -%>
     update-policy {
             grant rndc-key zonesub ANY;
--- a/templates/options.conf.erb
+++ b/templates/options.conf.erb
@@ -19,8 +19,8 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
 <% end -%>
-<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v4')) -%>
-listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_v4') %>; };
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on')) -%>
+listen-on port 53 { <%= scope.lookupvar('::dns::listen_on') %>; };
 <% end -%>
 <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::transfer_source')) -%>
 transfer-source <%= scope.lookupvar('::dns::transfer_source') %> <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::transfer_source_port')) %> port <%= scope.lookupvar('::dns::transfer_source_port') %><% end -%>;
--- a/templates/options.conf.erb
+++ b/templates/options.conf.erb
@@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
 <% end -%>
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
+listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
+<% end -%>
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::transfert_source')) -%>
+transfert-source <%= scope.lookupvar('::dns::transfert_source') %> <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::transfert_source_port')) %> port <%= scope.lookupvar('::dns::transfert_source_port') %><% end -%>;
+<% end -%>
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::notify_source')) -%>
+notify-source <%= scope.lookupvar('::dns::notify_source') %> <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::notify_source_port')) %> port <%= scope.lookupvar('::dns::notify_source_port') %><% end -%>;
+<% end -%>
 
 <% unless scope.lookupvar('::dns::allow_recursion').empty? -%>
 allow-recursion { <%= scope.lookupvar('::dns::allow_recursion').join("; ") %>; };
--- a/templates/options.conf.erb
+++ b/templates/options.conf.erb
@@ -19,8 +19,8 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
 <% end -%>
-<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
-listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v4')) -%>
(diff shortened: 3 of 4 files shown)


templates/options.conf.erb
```diff
@@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
 <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
 listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
 <% end -%>
+<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
+listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
```
domcleal: `port 53` should be optional, best perhaps to leave it out of both the directive and the parameter name in case it was later parameterised, so it's generally just "listen_on".

templates/named.zone.erb
```diff
@@ -6,9 +6,7 @@ zone "<%= @zone %>" {
     forwarders { <%= @forwarders.join('; ') %>; };
 <% end -%>
 <% end -%>
-<% if @manage_file == true -%>
```
domcleal: Why remove this?

manifests/init.pp
```diff
@@ -49,6 +50,11 @@
   $service_ensure       = $::dns::params::service_ensure,
   $service_enable       = $::dns::params::service_enable,
   $additional_options   = $::dns::params::additional_options,
+  $transfert_source     = $::dns::params::transfert_source,
```
domcleal: `transfert_source` or `transfer_source`? The definition in the params class is transfer_source, but I think the directive is without the 't'.

manifests/init.pp
```diff
@@ -49,6 +50,11 @@
...
+  $transfert_source_port= $::dns::params::transfert_source_port,
+  $notify_source        = $::dns::params::notify_source,
+  $notify_source_port   = $::dns::params::notify_source_port
+
```
domcleal: Remove the new line, and keep the trailing comma after notify_source_port please.

templates/options.conf.erb
ekohl: I agree on a generic ```listen_on```. That would be parallel to ```listen_on_v6```.

manifests/init.pp
```diff
@@ -49,6 +50,11 @@
   $service_ensure       = $::dns::params::service_ensure,
   $service_enable       = $::dns::params::service_enable,
   $additional_options   = $::dns::params::additional_options,
+  $transfert_source     = $::dns::params::transfert_source,
+  $transfert_source_port= $::dns::params::transfert_source_port,
```
ekohl: Would it be better if the user would provide the (optional) port as part of the transfer source?

jas01:
Le 12/12/2016 à 00:32:05-0800, Dominic Cleal a écrit
> @domcleal requested changes on this pull request.
>
> Please also check and fix the test failures. If you're able to add new tests
> that's appreciated, and squash down your commits.

Sorry for those problem. First time I use github/pullrequest.

>
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In templates/options.conf.erb:
>
> > @@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
>  <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
>  listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
>  <% end -%>
> +<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
> +listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
>
> port 53 should be optional, best perhaps to leave it out of both the directive
> and the parameter name in case it was later parameterised, so it's generally
> just "listen_on".
>

Well I agree with that but the point is bind accept lot of listen_on. Here
a example of that (pick up from bind9 source)

  listen-on {
          10/24;
          10.0.0.3;
  };

  listen-on port 53 { any; };

  listen-on { 5.6.7.8; };

  listen-on port 1234 {
          !1.2.3.4;
          1.2.3/24;
  };

so I think about a solution, find out if we want covert all case, it's
become very complexe.

So I though :

  not very lot of people use bind on another port than 53

so that's why I don't make a port number.

But if you think It's good idea I can do that.

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In templates/named.zone.erb:
>
> > @@ -6,9 +6,7 @@ zone "<%= @zone %>" {
>      forwarders { <%= @forwarders.join('; ') %>; };
>  <% end -%>
>  <% end -%>
> -<% if @manage_file == true -%>
>
> Why remove this?
>

Shit...it's definitively not a good github user....

Well I though that would be the next pull request.

The idea is :

  When you pass "manage_file" to false in a zone definition for me it's to
  tell the puppet/define not manage the file (not install the file), but
  that's not mean to put a zone config without the name of the file.

  Why ? I use bind with another system who create the zone file, so all I
  need is to tell puppet to create the configuration but not manage the
  file himself. So I need

zone "238.145.in-addr.arpa" {
    type master;
    file "/var/cache/bind/zones/db.145.238";
    update-policy {
            grant rndc-key zonesub ANY;
    };
}

but I don't want puppet manage the /var/cache/bind/zones/db.145.238

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In manifests/init.pp:
>
> > @@ -49,6 +50,11 @@
>    $service_ensure       = $::dns::params::service_ensure,
>    $service_enable       = $::dns::params::service_enable,
>    $additional_options   = $::dns::params::additional_options,
> +  $transfert_source     = $::dns::params::transfert_source,
>
> transfert_source or transfer_source? The definition in the params class is
> transfer_source, but I think the directive is without the 't'.

I'm french guy...so when I type in my keyboard I "automagically" add a «t».
Sorry. I change that.

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In manifests/init.pp:
>
> > @@ -49,6 +50,11 @@
>    $service_ensure       = $::dns::params::service_ensure,
>    $service_enable       = $::dns::params::service_enable,
>    $additional_options   = $::dns::params::additional_options,
> +  $transfert_source     = $::dns::params::transfert_source,
> +  $transfert_source_port= $::dns::params::transfert_source_port,
> +  $notify_source        = $::dns::params::notify_source,
> +  $notify_source_port   = $::dns::params::notify_source_port
> +
>
> Remove the new line, and keep the trailing comma after notify_source_port
> please.

Ok.

Regards.
--
Albert SHIH
DIO bâtiment 15
Observatoire de Paris
5 Place Jules Janssen
92195 Meudon Cedex
France

jas01:
Le 12/12/2016 à 00:32:05-0800, Dominic Cleal a écrit
> @domcleal requested changes on this pull request.
>
> Please also check and fix the test failures. If you're able to add new tests

So it's ok for the tests.

> that's appreciated, and squash down your commits.

Well...currently I'm not qualified for that.

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In templates/options.conf.erb:
>
> > @@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
>  <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
>  listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
>  <% end -%>
> +<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
> +listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
>
> port 53 should be optional, best perhaps to leave it out of both the directive
> and the parameter name in case it was later parameterised, so it's generally
> just "listen_on".

I got idea to change that. But I need to use each. Are you ok with that ?

>
> transfert_source or transfer_source? The definition in the params class is
> transfer_source, but I think the directive is without the 't'.

Fix in the last commit.

  https://github.com/jas01/puppet-dns/commit/e8432a1488d9b58473bce5cc627741a86f7637b3

> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In manifests/init.pp:
>
> > @@ -49,6 +50,11 @@
>    $service_ensure       = $::dns::params::service_ensure,
>    $service_enable       = $::dns::params::service_enable,
>    $additional_options   = $::dns::params::additional_options,
> +  $transfert_source     = $::dns::params::transfert_source,
> +  $transfert_source_port= $::dns::params::transfert_source_port,
> +  $notify_source        = $::dns::params::notify_source,
> +  $notify_source_port   = $::dns::params::notify_source_port
> +
>
> Remove the new line, and keep the trailing comma after notify_source_port
> please.

Done.

Regards.

ekohl:
> In templates/options.conf.erb:
> 
>  > @@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
>   <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_v6')) -%>
>   listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
>   <% end -%>
>  +<% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar('::dns::listen_on_53')) -%>
>  +listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
> 
>  port 53 should be optional, best perhaps to leave it out of both the directive
>  and the parameter name in case it was later parameterised, so it's generally
>  just "listen_on".
> Well I agree with that but the point is bind accept lot of listen_on. Here
> a example of that (pick up from bind9 source)
> 
>   listen-on {
>           10/24;
>           10.0.0.3;
>   };
> 
>   listen-on port 53 { any; };
> 
>   listen-on { 5.6.7.8; };
> 
>   listen-on port 1234 {
>           !1.2.3.4;
>           1.2.3/24;
>   };
> 
> so I think about a solution, find out if we want covert all case, it's
> become very complexe.
> 
> So I though :
> 
>   not very lot of people use bind on another port than 53
> 
> so that's why I don't make a port number.
> 
> But if you think It's good idea I can do that.

We already have the ```listen_on_v6``` that does the same as you describe for IPv6. It doesn't support any other port than 53, but it does allow multiple values. That's why I'd prefer if we simply have a ```listen_on``` variable.

jas01:
Le 15/12/2016 à 11:38:39-0800, Ewoud Kohl van Wijngaarden a écrit
>     In templates/options.conf.erb:
>
>         @@ -19,6 +19,15 @@ notify <%= scope.lookupvar('::dns::dns_notify') %>;
>         <% unless [false, nil, :undefined, :undef, ''].include?(scope.lookupvar
>         ('::dns::listen_on_v6')) -%>
>         listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
>         <% end -%>
>         +<% unless [false, nil, :undefined, :undef, ''].include?
>         (scope.lookupvar('::dns::listen_on_53')) -%>
>         +listen-on port 53 { <%= scope.lookupvar('::dns::listen_on_53') %>; };
>
>     port 53 should be optional, best perhaps to leave it out of both the
>     directive
>     and the parameter name in case it was later parameterised, so it's
>     generally
>     just "listen_on".
>     Well I agree with that but the point is bind accept lot of listen_on. Here
>     a example of that (pick up from bind9 source)
>
>     listen-on {
>     10/24;
>     10.0.0.3;
>     };
>
>     listen-on port 53 { any; };
>
>     listen-on { 5.6.7.8; };
>
>     listen-on port 1234 {
>     !1.2.3.4;
>     1.2.3/24;
>     };
>
>     so I think about a solution, find out if we want covert all case, it's
>     become very complexe.
>
>     So I though :
>
>     not very lot of people use bind on another port than 53
>
>     so that's why I don't make a port number.
>
>     But if you think It's good idea I can do that.
>
> We already have the listen_on_v6 that does the same as you describe for IPv6.

Not really because you can create multiple listen-on-v6, but I understand
your point of vue.

> It doesn't support any other port than 53, but it does allow multiple values.
> That's why I'd prefer if we simply have a listen_on variable.

Ok. I just change

  listen_on_53 to listen_on_v4

it's

  https://github.com/jas01/puppet-dns/commit/501c8d1cb97bd98c8079333a4de17ddd6d0c09b1

Regards

ekohl:
> Not really because you can create multiple listen-on-v6

That's true, there's a lot named can to that we don't support. We try to support the most common use cases because supporting all would most likely result in an unusable module. For more complex or non-standard deployments we do support custom config files (through concat).

> Ok. I just change listen_on_53 to listen_on_v4

I think I prefer ```listen_on``` since that maps to the bind config statement ```listen-on```, like ```listen_on_v6``` maps to ```listen-on-v6```.

jas01:
Le 16/12/2016 à 03:17:27-0800, Ewoud Kohl van Wijngaarden a écrit
>     Not really because you can create multiple listen-on-v6
>
> That's true, there's a lot named can to that we don't support. We try to
> support the most common use cases because supporting all would most likely
> result in an unusable module. For more complex or non-standard deployments we
> do support custom config files (through concat).
>
>     Ok. I just change listen_on_53 to listen_on_v4
>
> I think I prefer listen_on since that maps to the bind config statement

I understand.

> listen-on, like listen_on_v6 maps to listen-on-v6.

here :

  https://github.com/jas01/puppet-dns/commit/324bc10d7fee931281559f44a46d86d18b82d35b

Regards.

templates/named.zone.erb
ekohl: Could bring this part back?

jas01:
Le 16/12/2016 à 09:08:03-0800, Ewoud Kohl van Wijngaarden a écrit
> @ekohl requested changes on this pull request.
>
> I wonder how often the transfer_source_port and notify_source_port would be
> used, but I could accept them.

I used them ;-) That's why I add them.

>
> ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
>
> In templates/named.zone.erb:
>
> > @@ -6,9 +6,7 @@ zone "<%= @zone %>" {
>      forwarders { <%= @forwarders.join('; ') %>; };
>  <% end -%>
>  <% end -%>
> -<% if @manage_file == true -%>
>
> Could bring this part back?

Well....why ? I don't understand why you link those two thing together.

Or I don't understand the meaning of "manage_file", for me "manage_file"
meaning "Lets puppet manage the zone data header (zone.header.erb)". So put
manage_file to false meaning (for me) "I don't want puppet manage the zone
data header" that's not mean I don't want puppet manage the zone
definition.

As I say before :

  I have a application who create the zone data file. So I don't want
  puppet to create it. But I still want my zone definition :

    zone "myzone.fr" {
    type master;
    file "/var/cache/bind/zones/myzone.fr";
    update-policy {
            grant rndc-key zonesub ANY;
    };
    }

  but I don't want puppet create for me the

    /var/cache/bind/zones/myzone.fr

If it's not ok for you, maybe I can create a another parameter like
"create_file", what do you think ?

Regards

ekohl:
I'm going to close this for inactivity, but please let us know if you want to continue working on this.

PR #76, opened 2016-12-16
Document all class parameters

ekohl:
Closes GH-71

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -3,17 +3,91 @@
 # Install, configure and start dns service
 #
 # === Parameters:
+# $namedconf_path::        Path of the named config
+#                          type:Stdlib::Absolutepath
+#
+# $dnsdir::                Directory holding the named configs
+#                          type:Stdlib::Absolutepath
+#
+# $dns_server_package::    Name of the package to install
+#                          type:String
+#
+# $rndckeypath::           Path of the RNDC key
+#                          type:Stdlib::Absolutepath
+#
+# $optionspath::           Path of the named options
+#                          type:Stdlib::Absolutepath
+#
+# $publicviewpath::        Path of the config file holding all the zones
+#                          type:Stdlib::Absolutepath
+#
+# $vardir::                Directory holding the variable or working files
+#                          type:Stdlib::Absolutepath
+#
+# $namedservicename::      Name of the service
+#                          type:String
+#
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -3,17 +3,91 @@
...
+#                          type:Stdlib::Absolutepath
+#
+# $localzonepath::         File holding local zones like RFC1912 or RFC1918 files.
+#                          type:Optional[Stdlib::Absolutepath]
+#
+# $forward::               The forward option
+#                          type:Optional[String]
+#
+# $forwarders::            The forwarders option
+#                          type:Array[String]
```
stbenjam: I know we have a bunch of cases where we're expecting IP's, maybe we should get some better validation on them?

This looks promising: https://github.com/thrnio/puppet-ip

manifests/init.pp
```diff
@@ -3,17 +3,91 @@
...
+#
+# $listen_on_v6::          The listen-on-v6 option
+#                          type:Optional[String]
+#
+# $recursion::             The recursion option
+#                          type:Enum[yes, no]
```
stbenjam: I feel like these should be booleans that the templates convert to yes/no strings.

manifests/init.pp
ekohl: I think we could use ```Stdlib::Compat::Ipv4``` and ```Stdlib::Compat::Ipv6``` as well.

ekohl: According to http://www.zytrax.com/books/dns/ch7/queries.html#forwarders they can also include a port so simply forcing it to an IP is not the correct thing to do.

stbenjam: I suppose it's fine as is then, unless you wanted to create a type for DNS forwarders.

ekohl: Given this class isn't exposed in the installer that currently adds nothing.

stbenjam: It would once the variables themselves are typed, but I guess we can cross that bridge when we drop puppet 3.

ekohl: Exactly

mmoll:
merged, thanks @stbenjam  and @ekohl!

ekohl:
Lint now fails with the following error:
```
manifests/init.pp:126:WARNING: missing documentation for class parameter dns::additional_directives
```

I'll come up with a patch but the gate at the airport just opened.

PR #77, opened 2017-01-02
Add allow_query parameter for zones

dlucredativ:


domcleal:
Merged, thanks @dlucredativ.

PR #78, opened 2017-01-06
Add view support

dlucredativ:
This PR enables zones to be put and reconfigured in one or multiple BIND views.


ekohl:
Why is the parser function needed? Can't ```dns::view``` include the concat fragment? ```dns::zone``` does it that way.

ekohl:
I'm on vacation for the next week. Maybe some other contributor could have a look, but otherwise I'll get back to you later on this.

ekohl:
I wonder if the same can be achieved by a ```dns::zone::view``` define which does the same as the function. Currently I'm having a hard time to understand what it would exactly result into and I'm afraid that might turn it into a maintenance burden.

dlucredativ:
> I wonder if the same can be achieved by a dns::zone::view define which does the same as the function. Currently I'm having a hard time to understand what it would exactly result into and I'm afraid that might turn it into a maintenance burden.

Without lambdas, the only way (that I know) to create resources from an array, is either:
1. pass an array of titles to a resource declaration, or
2. create a special hash structure and feed it to create_resources().

As long as I want $title == $zone, method 1 is good enough, but for use case 2 it is not suitable. And if I am not mistaken, method 2 is of no help either, because I cannot create that hash structure without some loop statement in the first place.

If you really don't like the parser function, target_views needs to become target_view that does accept only one view. Users of the module would need to iterate over an array for themselves. I am fine with it, I have future parser enabled.

ekohl:
@mmoll what do you think?

ekohl:
I've had a bit deeper look at how bind handles views. The biggest factor that makes it complex that if you use a view, everything has to be in views. That brings in a lot of complexity that would not be here otherwise. I'm thinking if there is a way to not be so invasive in the existing zones while still supporting views, but that may be wishful thinking. Let me get back to you on that.

dlucredativ:
Yes, you are correct. This is why I introduced the use_views flag as well as the warning inside dns::zone not to try to mix both. There may be some future BIND versions which suport this, so I did not make it a fail().

aptituz:
Hey @ekohl, any news on this? :)

ekohl:
And my apologies for taking a long time to review it.

dlucredativ:
Hi @ekohl, thanks for the review! I will comment on the changes above.

manifests/zone.pp
```diff
@@ -14,46 +15,59 @@
     $allow_transfer = [],
     $allow_query    = [],
     $also_notify    = [],
-    $zone           = $title,
-    $contact        = "root.${title}.",
+    $zone           = undef,
+    $contact        = undef,
     $zonefilepath   = $::dns::zonefilepath,
-    $filename       = "db.${title}",
+    $filename       = undef,
```
dlucredativ: I think you are right. One does not want a shared zonefile for a zone in multiple views anyway. Bind will rename such a zonefile to something like db-xyzabcd.

manifests/zone.pp
```diff
@@ -14,46 +15,59 @@
...
-  validate_array($masters, $allow_transfer, $allow_query, $forwarders, $also_notify)
+  validate_array($masters, $allow_transfer, $allow_query, $forwarders, $also_notify, $target_views)
   validate_re($forward, '^(first|only)$', 'Only \'first\' or \'only\' are valid values for forward field')
   if $dns_notify {
     validate_re($dns_notify, '^(yes|no|explicit)$', 'Only \'yes\', \'no\', or \'explicit\' are valid values for dns_notify field')
   }
 
-  $zonefilename = "${zonefilepath}/${filename}"
+  if $zone     == undef { $_zone = $title } else { $_zone = $zone }
+  if $contact  == undef { $_contact  = "root.${_zone}." } else { $_contact = $contact }
```
dlucredativ: ok

manifests/zone.pp
```diff
@@ -14,46 +15,59 @@
...
   if $manage_file {
-    file { $zonefilename:
-      ensure  => file,
-      owner   => $dns::user,
-      group   => $dns::group,
-      mode    => '0644',
-      content => template('dns/zone.header.erb'),
-      replace => false,
-      notify  => Service[$::dns::namedservicename],
+    unless defined(File[$zonefilename]) {
```
dlucredativ: Yes, as mentioned above, zonefiles should be unique anyway, so we could drop it.

dlucredativ: Hm, maybe we cannot drop it, or I need at least a hint considering the following use case:
- define a master dns::zone *master_zone*
- define custom content for the zonefile for *master_zone* outside of ::dns

To do that, I need to have `manage_file => true` because of
```
<% if @manage_file == true -%>
    file "<%= @zonefilename %>";
<% end -%>
```
If we drop the `unless defined(File[$zonefilename])` the content attribute is set in dns::zone, and I am not able to override it in my own Puppet code.

dlucredativ:
Hi, I have included some change requests and rebased against the current master.

manifests/zone.pp
ekohl: I think I'd rather have an explicit parameter than rely on ```defined```. You're right that ```manage_file``` is already used in another place and maybe we need to change that. I'm still thinking about the best solution here. At the very least you could combine the if statement into ```if $manage_file and not defined(File[$zonefilename]) {``` for a smaller diff.

templates/named.zone.erb
```diff
@@ -30,3 +30,4 @@ zone "<%= @zone %>" {
     notify <%= @_dns_notify %>;
 <% end -%>
 };
+
```
ekohl: Why is this empty line needed?

dlucredativ: Probably unintentional, but I will check tomorrow how the config looks like without the newline, and remove it then.

dlucredativ: Removed it.

manifests/zone.pp
dlucredativ: For now, I combined the two conditionals as you requested. Another, more flexible solution might be a new parameter `$zonefilecontent`, which defaults to undef. If undef, use the current template, if defined, use `$zonefilecontent` instead.

ekohl:
Given we dropped puppet 3 support, can you rewrite it using Puppet 4 code? I believe it can be simplified then.

dlucredativ:
Hm, I made a parser-function-free version based on the current master which will probably work with Puppet 4.x. However, I have trouble with Stdlib::Absolutepath on future-parser-enabled 3.8.5 puppet master. :/

manifests/view.pp
```diff
@@ -0,0 +1,60 @@
...
+    $forward              = 'first',
+    $recursion            = 'yes',
+    $dnssec_enable        = 'yes',
+    $dnssec_validation    = 'yes',
+    $dns_notify           = 'yes',
+    $include_localzones   = true,
+    $include_defaultzones = true,
+    $order                = '-',
+) {
+  validate_array($match_clients, $match_destinations, $allow_transfer, $allow_recursion, $allow_query, $allow_query_cache, $also_notify, $forwarders)
```
ekohl: Could you rewrite these to types as well?

templates/named.view.erb
```diff
@@ -0,0 +1,44 @@
+view "<%= @view %>" {
```
ekohl: Is this template still used?

templates/named.view_header.erb
```diff
@@ -0,0 +1,55 @@
...
+<% if @dns_notify -%>
+    notify <%= @dns_notify %>;
+<% end -%>
+<% if @match_recursive_only -%>
+    match-recursive-only <%= @match_recursive_only %>;
+<% end -%>
+<% if @recursion -%>
+    recursion <%= @recursion %>;
+<% end -%>
+<% if @dnssec_enable -%>
```
ekohl: This is verified to be ```yes|no``` so this is always true. There are a few other parameters that don't need an if statement.

templates/named.view_header.erb
```diff
@@ -0,0 +1,55 @@
...
+<% end -%>
+<% if @dnssec_validation -%>
+    dnssec-validation <%= @dnssec_validation %>;
+<% end -%>
+<% if @notify -%>
+    notify <%= @notify %>;
+<% end -%>
+
+<% if @include_localzones -%>
+<% unless [nil, :undefined, :undef, 'unmanaged'].include?(scope.lookupvar("::dns::localzonepath")) -%>
```
ekohl: Should this be a check in the class? Feels to me like this is an invalid configuration.

manifests/view.pp
dlucredativ: Done.

templates/named.view.erb
dlucredativ: No, removed it.

templates/named.view_header.erb
dlucredativ: Right, I removed some of the if blocks.

templates/named.view_header.erb
dlucredativ: Well, I am not satisfied with that either. $dns::localzonepath has a non-undef default value that one cannot unset via a parameter. However, I want to provide the possibility to manage RFC1912/RFC1918 zones via dns::zone (with and without `$enable_views`), in which case I don't want to include `$localzonepath`. Since `97b99cbf` until `fa16bad`, it was possible to achieve this by setting the `localzonepath` parameter to the empty string. Now, the empty string is not allowed any more, as it is not a valid Absolutepath. So I introduced a Variant Enum value called "unmanaged" instead. Maybe you can think of a more elegant way?

dlucredativ:
> Could you show the errors you are getting?

I'm getting `Expected parameter 'namedconf_path' of 'Class[Dns]' to have type Stdlib::Absolutepath`. It seems that Puppet 3.8.5 does not understand custom data types inside the `types`folder of a module. I will have to use a fork without external types then.

manifests/init.pp
```diff
@@ -31,6 +31,13 @@
 #                               type:Stdlib::Absolutepath
 #
 # $localzonepath::              File holding local zones like RFC1912 or RFC1918 files.
+#                               The special value 'unmanaged' can be used if one plans
+#                               to create custom RFC1912/RFC1918 zones via ::dns,
+#                               where the inclusion of package-shipped zone files is
+#                               not desired.
+#                               type:Optional[Variant[Enum['unmanaged'],Stdlib::Absolutepath]]
```
ekohl: These types are no longer needed and were removed in https://github.com/theforeman/puppet-dns/commit/87df08cd842a5465caa404d1dabe1213e8218020 (which is also causing a merge conflict.

dlucredativ:
> Previously I wrote an acceptance test in ekohl@fcc6442 and it's mostly up to date. The before block at https://github.com/theforeman/puppet-dns/blob/87df08cd842a5465caa404d1dabe1213e8218020/spec/acceptance/dns_spec.rb#L4-L13 should be used instead. Would you mind including that in the PR?

Which before-block do you want to have, the one with or without `package_name = 'bind9'`?

ekohl:
> Which before-block do you want to have, the one with or without package_name = 'bind9'?

The one without.

manifests/init.pp
```diff
@@ -22,6 +22,12 @@
 # $zonefilepath::               Directory containing zone files
 #
 # $localzonepath::              File holding local zones like RFC1912 or RFC1918 files.
+#                               The special value 'unmanaged' can be used if one plans
+#                               to create custom RFC1912/RFC1918 zones via ::dns,
+#                               where the inclusion of package-shipped zone files is
+#                               not desired.
+#
+# $defaultzonepath::            File holding default zone includes like db.local.
```
ekohl: This is in params.pp but not exposed in init.pp. It should either be exposed or removed from the documentation. Currently the isn't managed either so it doesn't exist. That causes the acceptance test to fail (named won't start up on EL7).

manifests/zone.pp
```diff
@@ -24,21 +25,45 @@
...
+      order   => "${view}-11-${zone}-1",
+    }
+
+    unless ($view == '_GLOBAL_' or defined(Dns::View[$view])) {
+      fail("Please define a dns::view '${view}' before using it as a dns::zone target")
+    }
   }
 
-  if $manage_file {
+  if $manage_file and !defined(File[$zonefilename]) {
```
ekohl: I'm still not happy about the use of defined here. How about adding a parameter ```use_file``` (or something like that) and changing the logic in ```named.zone.erb``` to ```<% if @manage_file || @use_file -%>```? Then ```use_file``` can default to false and get the same behaviour but allow for the desired behaviour.

manifests/init.pp
dlucredativ: Hmpf, I broke this one during a merge conflict resolution, sorry.

manifests/zone.pp
dlucredativ: Hi, I implemented your idea via `manage_file_name`.

dlucredativ:
@ekohl fyi, I will be unavailable next week.

ekohl:
And I'm unavailable until next week. I'll do some testing then and if all goes well I'll merge it

dlucredativ:
Hi @ekohl , any news?

ekohl:
I was looking into this last week but got distracted by another issue that I've almost tackled. My apologies for taking so long.

ekohl:
Thanks @dlucredativ! I merged it with some changes:
* Squashed into a single commit
* Fixed a minor merge conflict introduced by my style changes commit
* Only create the views directory if views are enabled to reduce impact for users without views

PR #79, opened 2017-01-15
add additional_directives parameter for top-scope bind directives in named.conf

dlucredativ:
As discussed in https://github.com/theforeman/puppet-dns/pull/72 here comes a version with a loop inside the template which injects the `additional_directives` content into named.conf.

### DIFF
--- a/templates/named.conf.erb
+++ b/templates/named.conf.erb
@@ -23,6 +23,12 @@ acl "<%= acl_name %>"  {
         <%- end -%>
 };
 <%- end -%>
+<%- if scope.lookupvar('::dns::additional_directives').any? -%>
+// additional directives
+<%- scope.lookupvar('::dns::additional_directives').each do |directive| -%>
+<%= directive %>
+<%- end -%>
+<%- end -%>
 
 // Public view read by Server Admin
 include "<%= scope.lookupvar('::dns::publicviewpath') %>";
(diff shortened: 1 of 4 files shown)


templates/named.conf.erb
```diff
@@ -24,5 +24,12 @@ acl "<%= acl_name %>"  {
 };
 <%- end -%>
 
+<%- if scope.lookupvar('::dns::additional_directives').any? -%>
+// additional directives
+<% scope.lookupvar('::dns::additional_directives').each do |directive| -%>
+<%= directive %>
+<%- end -%>
+<%- end -%>
```
ekohl: I realize that other blocks don't do this either, but I wonder if it should include the empty line so it won't add an empty line for users who don't use this directive. Then it would look like this:

```erb
<%- end -%>

<%- end -%>
// Public view read by Server Admin
```

dlucredativ: The spec test in dns_init_spec.rb line 36 still goes through, so there will be no change for users after the module update.    Is that what you mean?

ekohl: The test filters out empty lines which is a bit of a simplification for us developers, but users might notice it.

dlucredativ: You are right, I updated the PR source after some apply tests. Now, without additional_directives, the named.conf stays as it was.

mmoll:
I'm not a big fan of such blanco-parameters, but it seems to work well.

mmoll:
merged, thanks @dlucredativ!

PR #81, opened 2017-03-03
replace all validate functions with datatypes

bastelfreak:


bastelfreak:
I recommend bumping puppet not to 4.0.0 but to 4.6.1. This is the recommended version by Puppet, Vox Pupuli agreed to use this as well. Let me know if I should do any more validations or other stuff in this PR.

manifests/init.pp
```diff
@@ -101,53 +101,35 @@
...
-  $dnssec_validation     = $::dns::params::dnssec_validation,
-  $namedconf_template    = $::dns::params::namedconf_template,
-  $acls                  = $::dns::params::acls,
-  $optionsconf_template  = $::dns::params::optionsconf_template,
-  $controls              = $::dns::params::controls,
-  $service_ensure        = $::dns::params::service_ensure,
-  $service_enable        = $::dns::params::service_enable,
-  $additional_options    = $::dns::params::additional_options,
-  $additional_directives = $::dns::params::additional_directives,
+  $namedconf_path                                                 = $::dns::params::namedconf_path,
```
ekohl: In the documentation we have a type for every parameter. Could you add those as well?

manifests/init.pp
```diff
@@ -101,53 +101,35 @@
...
+  Array[String] $allow_query                                      = $::dns::params::allow_query,
+  $empty_zones_enable                                             = $::dns::params::empty_zones_enable,
+  Optional[Enum['yes', 'no', 'explicit']] $dns_notify             = $::dns::params::dns_notify,
+  Enum['yes', 'no'] $dnssec_enable                                = $::dns::params::dnssec_enable,
+  Enum['yes', 'no', 'auto'] $dnssec_validation                    = $::dns::params::dnssec_validation,
+  $namedconf_template                                             = $::dns::params::namedconf_template,
+  Hash $acls                                                      = $::dns::params::acls,
+  $optionsconf_template                                           = $::dns::params::optionsconf_template,
+  Hash $controls                                                  = $::dns::params::controls,
+  Enum['running', 'true', 'stopped', 'false'] $service_ensure     = $::dns::params::service_ensure,
```
ekohl: It feels like here should be a common type for this.

bastelfreak:
@ekohl I fixed the documentation and added validation for all types.

ekohl:
I think we can already merge the documentation fix and stdlib bump. The rest may require some branching.

bastelfreak:
sure. I can cherry-pick them and provide them in a separate PR.

mmoll:
@bastelfreak Please rebase onto current master and bump the major version and minimum puppet version in metadata.json. Also, squashing all commits together would be appreciated.

manifests/init.pp
```diff
@@ -101,53 +101,35 @@
...
+  Array[String] $allow_query                                      = $::dns::params::allow_query,
+  Enum[yes, no] $empty_zones_enable                               = $::dns::params::empty_zones_enable,
+  Optional[Enum['yes', 'no', 'explicit']] $dns_notify             = $::dns::params::dns_notify,
+  Enum['yes', 'no'] $dnssec_enable                                = $::dns::params::dnssec_enable,
+  Enum['yes', 'no', 'auto'] $dnssec_validation                    = $::dns::params::dnssec_validation,
+  String $namedconf_template                                      = $::dns::params::namedconf_template,
+  Hash[String, Array[String]] $acls                               = $::dns::params::acls,
+  String $optionsconf_template                                    = $::dns::params::optionsconf_template,
+  Hash[String, Hash[String, Data]] $controls                      = $::dns::params::controls,
+  Enum['running', 'true', 'stopped', 'false'] $service_ensure     = $::dns::params::service_ensure,
```
mmoll: `manifests/init.pp:128:WARNING: quoted boolean value found` but I'm unsure how to handle this best...

bastelfreak: Personally I would prefer it to only match for 'running' and 'stopped'. As an alternative we could do:

```puppet
Variant[Enum['running', 'stopped'], Boolean]

mmoll: I'd like to go with the second variant, as per https://docs.puppet.com/puppet/latest/types/service.html#service-attribute-ensure booleans are still valid.

bastelfreak:
@mmoll rebased, squashed, bumped versions.

---

I also noticed that there is still support for the legacy concat version 1, which works with defined resources. Are you interested in a PR which bumps it to the new major release which uses native types? Since the next release of puppet-dns has to be a major one it might be a good idea to bump concat now and not later?

mmoll:
merged, danke @bastelfreak!

> Are you interested in a PR which bumps it to the new major release which uses native types?

sure 👍

PR #82, opened 2017-03-04
Doc updates + bump stdlib

bastelfreak:
these two got cherrypicked from https://github.com/theforeman/puppet-dns/pull/81

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -34,13 +34,13 @@
 #                               type:Optional[Stdlib::Absolutepath]
 #
 # $forward::                    The forward option
-#                               type:Optional[String]
+#                               type:Optional[Enum['only', 'first']]
 #
 # $forwarders::                 The forwarders option
 #                               type:Array[String]
 #
 # $listen_on_v6::               The listen-on-v6 option
-#                               type:Optional[String]
+#                               type:Optional[Variant[String, Boolean]]
 #
 # $recursion::                  The recursion option
 #                               type:Enum[yes, no]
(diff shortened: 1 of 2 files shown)


bastelfreak:
eh wat? I will take a closer look tomorrow again.

bastelfreak:
nevermind. I've never seen $thing parsing the documentation and blowing up on broken braces :dancer:

ekohl:
We use puppet-lint-param-docs for that since our installer uses the docs (and their types) to provide help and validation.

manifests/init.pp
```diff
@@ -34,13 +34,13 @@
 #                               type:Optional[Stdlib::Absolutepath]
 #
 # $forward::                    The forward option
-#                               type:Optional[String]
+#                               type:Optional[Enum['only', 'first']]
```
ekohl: As someone still a bit new to types: is there a difference between quoted and unquoted values here?

bastelfreak: I'm not sure, I've never tested them without quotes. But since they are strings I would expect them to be quoted.

ekohl: I'm mostly asking because we have some others below where they aren't quoted. I'd prefer using quoted only but until I saw both versions I hadn't thought about it that much.

bastelfreak: since the introduction of types requires a major version bump we could just switch all to quoted?

domcleal: Either's valid, in the same way you can use bare words in a manifest for single world strings. I'd prefer quoted.

ekohl:
Thanks

PR #86, opened 2017-04-09
migrate to native concat type+provider

bastelfreak:
as discussed in https://github.com/theforeman/puppet-dns/pull/81

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -2,34 +2,44 @@
 class dns::config {
   group { $dns::params::group: }
 
-  concat { $::dns::publicviewpath:
+  concat_file { $::dns::publicviewpath:
     owner => root,
     group => $dns::params::group,
     mode  => '0640',
+    tag   => 'concat_file_publicviewport',
   }
 
-  concat::fragment { 'dns_zones+01-header.dns':
+  concat_fragment { 'dns_zones+01-header.dns':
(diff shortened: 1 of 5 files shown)


domcleal:
Is this necessary? The defined types still use the native types underneath. My understanding is that the defined types remain the public API for using the module.

bastelfreak:
I talked to the puppetlabs people on the IRC. The native types and providers are also considered as public API. concat 2.X should be used, instead of the 1.X versions, because it uses the more efficient types and providers. I thought it's a good idea to switch to the T&P directly and not only bumping the concat version in metadata.json, because now it's more clear that the T&P are actually used.

domcleal:
2.x's definitely superior, but I don't think using the types directly is an improvement. It looks more difficult (needing to specify tags), and lacks some of the checks that's in the defined types.

Good to know that the types remain part of the public API, but the documented usage is still the easier defined types. I remain unsure that this is an improvement.

ekohl:
I agree with @domcleal and don't see this as an improvement. As there's no real improvement when using pl-concat 2.x I'd prefer to keep the current code.

ekohl:
I'm leaning to closing this.

domcleal:
Closing, please re-open if you think it's necessary or an improvement, but as it stands I think it's best to continue using the defined types on all versions.

PR #87, opened 2017-04-11
Remove duplicated data type from parameter documentation

domcleal:
Corrects two parameters to the more specific type from the docs, and
permits any data in `additional_options` to match usage in existing
tests.


mmoll:
merged, thanks @domcleal!

PR #97, opened 2017-09-29
Use contain for chaining

ekohl:
This prevents a cyclic dependency when using the zones parameter.


mmoll:
merged, bedankt @ekohl!

PR #103, opened 2017-11-24
Fixes #21756 - Make rndc hash algorithm configurable

witlessbird:


witlessbird:
This PR removes "automatic" rndc key configuration, with one that is more verbose, but appears to be compatible with the rest of the configuration created by this module. Please note that newer versions (starting with 9.10.0) of rndc-confgen support multiple hash functions, including SHA512.

witlessbird:
Both bind and dhcpd use isc's implementations of crypto hash functions (including MD5) and appear to be unaffected by openssl operating in FIPS mode. I don't think this PR is needed.

Issue #94, opened 2017-08-04
Add support for adding keys for nsupdate


mstinsky:
Hey,

I am in the middle of evaluating a foreman puppet setup, trying out all the foreman smart proxy features. I want to do everything through puppet and u guys have pretty good modules already for that. But i can't find how I would be able to add a key to bind for foremans nsupdates, shoulnd this be supported by this module or am I missing something ?

Best regards

ekohl:
That's a good question I didn't really consider. When I still used bind + dhcpd the services were on the same machine so limiting to localhost was sufficient for me. It looks like this is only doable with custom includes and not trivial out of the box.

Pull requests are highly appreciated but a description of what you'd expect is also very useful.

mstinsky:
I will look into if i can make a pull request happen for that.
What i expect as a minimum for this is to do the exact steps that are pointed out in the Foreman Manual (4.3.5.2 BIND):

1. Add a key to the named.conf.
2. Add update-policy for that key to specified zone.
Optional but would be a big plus: Create the key through this module.

ekohl:
Let me know if you want help. Also very early work in progress reviews are not a problem.

PR #108, opened 2018-02-11
Adds control keys and specifying update policy

zyronix:
Fixes GH-94
This commit adds a new resource type key, which creates a rndc key which can be used to control bind using nsupdate.
Also this commit allows to specify update policy rules for zones.

I see that the beaker tests are failing and I also have not written a beaker test for this one.
Please let me know a beaker test is required.

### DIFF
--- a/manifests/key.pp
+++ b/manifests/key.pp
@@ -1,0 +1,42 @@
+# Generate a new key for the dns
+#
+# === Parameters:
+#
+# $secret::                     This is the secret to be place inside the keyfile, if left empty the key will be generated
+#    
+define dns::key(
+  String               $algorithm    = 'hmac-md5',
+  String               $filename     = "${name}.key",
+  Optional[String]     $secret       = undef,
+  Stdlib::Absolutepath $keydir       = $::dns::dnsdir,
+  Integer              $keysize      = 512,
+) {
+  $keyfilename = "${keydir}/${filename}"
+
+  if $secret {
+    file {$keyfilename:
+      ensure  => file,
+      owner   => $dns::user,
+      group   => $dns::group,
+      mode    => '0640',
+      content => template('dns/key.erb'),
+      notify  => Service[$::dns::namedservicename],
+    }
+  } else {
+    exec { "create-${filename}":
+      command => "${dns::rndcconfgen} -r /dev/urandom -a -c ${keyfilename} -b ${keysize} -k ${name}",
+      creates => $keyfilename,
+      notify  => Service[$::dns::namedservicename],
+    }-> file { $keyfilename:
+      owner => 'root',
+      group => $dns::params::group,
+      mode  => '0640',
+    }
+  }
+
+  concat::fragment { "named.conf+20-key-${name}.dns":
+    target  => $::dns::namedconf_path,
+    content => "include \"${keyfilename}\";\n",
+    order   => '20',
+  }
+}
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -6,31 +6,37 @@
 #
 # $manage_file_name::           Whether to set the file parameter in the zone file.
 #
+# $update_policy_rules::        This can be used to specifiy additional update policy rules in the following format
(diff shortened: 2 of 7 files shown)


manifests/key.pp
```diff
@@ -0,0 +1,43 @@
...
+#    
+define dns::key(
+  String               $algorithm    = 'hmac-md5',
+  String               $filename     = "${name}.key",
+  Optional[String]     $secret       = undef,
+  Stdlib::Absolutepath $keydir       = $::dns::dnsdir,
+) {
+  $keyfilename = "${keydir}/${filename}"
+
+  if ! $secret {
```
ekohl: This can be ```unless $secret``` but it might be better to use ```if $secret { file ... } else { exec ... file ... }``` instead

manifests/key.pp
```diff
@@ -0,0 +1,43 @@
...
+    exec { "create-${filename}":
+      command => "${dns::rndcconfgen} -r /dev/urandom -a -c ${keyfilename} -b 512 -k ${name}",
```
ekohl: Would it make sense to make the keysize a parameter?

manifests/key.pp
```diff
@@ -0,0 +1,43 @@
...
+      ensure  => file,
+      owner   => $dns::user,
+      group   => $dns::group,
+      mode    => '0640',
+      content => template('dns/key.erb'),
+      notify  => Service[$::dns::namedservicename],
+    }
+  }
+
+  concat::fragment { 'named.conf+20-keys.dns':
```
ekohl: You should include the key name in here because otherwise you get duplicate resource definitions when ```dns::key``` is used multiple times.

templates/named.zone.erb
```diff
@@ -12,6 +12,11 @@ zone "<%= @zone %>" {
 <% if @zonetype == 'master' -%>
     update-policy {
             grant rndc-key zonesub ANY;
+            <%- unless @update_policy_rules.empty? -%>
```
ekohl: I think this line is redundant since iterating over an empty hash will not produce any output either.

spec/defines/dns_key_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      verify_concat_fragment_exact_contents(catalogue, 'named.conf+20-keys.dns', [
+        'include "/etc/foreman_key.key";',
+      ])
+
+    end
+  end
+end
+
+
+#   # context 'secret set' do
```
ekohl: What's the plan with this commented code?

templates/key.erb
```diff
@@ -0,0 +1,4 @@
+key "<%= @name %>" {
+    algorithm <%= @algorithm %>;
+    secret "<%= @secret %>";
+};
```
ekohl: Could yuo add a newline here?

manifests/zone.pp
```diff
@@ -6,31 +6,37 @@
...
+  String $zone                                          = $title,
+  Optional[String] $contact                             = undef,
+  Stdlib::Absolutepath $zonefilepath                    = $::dns::zonefilepath,
+  String $filename                                      = "db.${title}",
+  Boolean $manage_file                                  = true,
+  Boolean $manage_file_name                             = false,
+  Enum['first', 'only'] $forward                        = 'first',
+  Array $forwarders                                     = [],
+  Optional[Enum['yes', 'no', 'explicit']] $dns_notify   = undef,
+  Hash[String, Hash[String, Data]] $update_policy_rules = {},
```
ekohl: Instead of ```Data``` you could add a custom data type to verify the fields in it similar to https://github.com/theforeman/puppet-foreman/pull/601 but that does break with kafo so I'd be fine with holding that off.

zyronix:
Bedankt!. That was a very fast reply and review 👍 . I have applied all the requested changes. Do you want me to squash both commits?

ekohl:
I think we already had failing beaker tests before this patch so I think you can ignore that. Especially since we don't test the modified code.

tiernap:
oh snap! looks like i created a pull request for the same issue. #109

zyronix:
What is next? Do I have to do something to get this merged?

mmoll:
merged, bedankt @zyronix!

PR #130, opened 2018-11-22
feat: add keys parameter and create_resources accordingly

marcdeop:
Thanks to #108 we can use custom keys :+1: 

We wanted to take it a step further and use the same method for the keys that exists for the zones: `create_resources` via a hash.

This PR achieves this.

Thanks to [Erasys GmbH](http://www.erasys.de/) for their support


ekohl:
Thanks!

PR #139, opened 2019-04-19
Make managing BIND system group optional

antaflos:
Add class parameter dns::group_manage which defaults to true, so that
the module keeps managing the system group for BIND ("bind" or "named"),
but allow the user to set it to false. This way the group can be managed
outside of this module's scope.

Fixes GH-140

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,7 +1,9 @@
 # Configure dns
 # @api private
 class dns::config {
-  group { $dns::params::group: }
+  if $dns::group_manage {
+    group { $dns::params::group: }
+  }
 
   concat { $dns::publicviewpath:
     owner => root,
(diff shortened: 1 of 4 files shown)


manifests/config.pp
```diff
@@ -1,7 +1,9 @@
 # Configure dns
 # @api private
 class dns::config {
-  group { $dns::params::group: }
+  if $dns::group_manage == true {
```
ekohl:
```suggestion
  if $dns::group_manage {
```

antaflos: @ekohl Thanks for the comment! I have updated the PR and removed the explicit comparison.

antaflos:
Gently bumping this :)

mmoll:
merged, thanks @antaflos!

PR #142, opened 2019-05-08
Don't set forbidden zone options for zone type 'forward'

### LINKED ISSUE #141: Module can generate invalid configuration when zonetype => forward
When declaring a `dns::zone` resource with `zonetype => 'forward'` and also setting `manage_file`/`manage_file_name`, or even `masters`, `allow_transfer`, etc then the resulting configuration in `zones.conf` is invalid and named refuses to start, with error messages like this:

```
named[13888]: /etc/bind/zones.conf:14: option 'allow-query' is not allowed in 'forward' zone 'foo.example.com'
named[13888]: /etc/bind/zones.conf:13: option 'allow-transfer' is not allowed in 'forward' zone 'foo.example.com'
named[13888]: /etc/bind/zones.conf:15: option 'also-notify' is not allowed in 'forward' zone 'foo.example.com'
named[13888]: /etc/bind/zones.conf:11: option 'masters' is not allowed in 'forward' zone 'foo.example.com'
named[13888]: /etc/bind/zones.conf:12: option 'notify' is not allowed in 'forward' zone 'foo.example.com'
named[13888]: loading configuration: failure
named[13888]: exiting (due to fatal error)
```

I will submit a PR to fix this.

antaflos:
When zonetype => forward is set for a dns::zone resource then ignore any
forbidden options that might be set as parameters when rendering the
named.zone.erb template. Such forbidden options are "file", "masters",
"allow-transfer", "allow-query", "notify" and "also_notify".

This prevents the module from generating an invalid named configuration
(in zones.conf) that would result in named failing to (re)start.

Fixes #141


antaflos:
In this case I thought it may be easier and more efficient to make the template aware of the zone type and let it "do the right thing" on its own, without failing the catalog compilation.

The use case I thought of was this: when most zones are slave zones we can set defaults for the `dns::zone` type in Hiera or in a profile, like `zonetype => slave` and `manage_file_name => true`. Then we can override the `zonetype` default for specific zones by setting `zonetype => forward` and we don't have to set or change `manage_file_name`; it can remain at its set default value since it is just ignored for the forward zone type.

But if you want I can also change the behaviour like you suggested and fail with a hard error if this is the preferable way to go for this module.

antaflos:
Gently bumping this as well :)

ekohl:
Thanks!

PR #143, opened 2019-05-08
Allow setting service restart command

antaflos:
By default the service resource that manages the BIND service will
restart the service when notified by configuration changes. This can be
disruptive.

This change allows the user to set a custom restart command via
`dns::service_restart_command` for the service resource, such as
`/usr/sbin/service bind9 reload` or `/bin/systemctl reload bind9.service`
or even `/usr/sbin/rndc reload`, to avoid restarting the whole named
process. By default no restart command is set.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,6 +93,14 @@
 #   The ensure attribute on the service
 # @param service_enable
 #   Whether to enable the service (start at boot)
+# @param service_restart_command
+#   Custom command to use when the service will be restarted (notified by
+#   configuration changes). Will be passed directly to the restart parameter of
+#   the contained service resource. This is useful when you want BIND to reload
+#   its configuration instead of restarting the whole process, for example by
+#   setting `service_restart_command` to `/usr/sbin/service bind9 reload` or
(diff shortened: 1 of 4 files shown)


antaflos:
I have rebased the branch on the current master and force-pushed.

manifests/init.pp
```diff
@@ -93,6 +93,14 @@
...
 # @param service_enable
 #   Whether to enable the service (start at boot)
+# @param service_restart_command
+#   Custom command to use when the service will be restarted (notified by
+#   configuration changes). Will be passed directly to the restart parameter of
+#   the contained service resource. This is useful when you want BIND to reload
+#   its configuration instead of restarting the whole process, for example by
+#   setting `service_restart_command` to `/usr/sbin/service bind9 reload` or
+#   `/usr/sbin/rndc reload`. Default 'undef' so the service resource default is
+#   used.
```
ekohl: Another example can be `/usr/bin/systemctl try-reload-or-restart bind9` (but I like the `undef` default, please keep that :))

antaflos: I have added your example to the parameter documentation!

ekohl:
Thanks!

PR #144, opened 2019-05-08
Validate named.conf and zones.conf using named-checkconf

antaflos:
Make use of the `validate_cmd` parameter in concat to run
named-checkconf on the target file. This way the validity of to-be-updated
configuration files named.conf or zones.conf is checked before actually
writing the new configuration and restarting the named service. This
prevents named from loading invalid configuration settings that would
result in named failing to start.

The file named.conf.options itself cannot be checked with
named-checkconf because its content is only valid inside the
"options { };" directive.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -6,9 +6,10 @@ class dns::config {
   }
 
   concat { $dns::publicviewpath:
-    owner => root,
-    group => $dns::params::group,
-    mode  => '0640',
+    owner        => root,
+    group        => $dns::params::group,
+    mode         => '0640',
+    validate_cmd => "${dns::named_checkconf} %",
   }
 
   if $dns::enable_views {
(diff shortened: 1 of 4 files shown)


antaflos:
This needs more work to appease the acceptance tests.

antaflos:
I have rebased the branch on the current master and force-pushed.

manifests/config.pp
```diff
@@ -6,9 +6,10 @@
   }
 
   concat { $dns::publicviewpath:
-    owner => root,
-    group => $dns::params::group,
-    mode  => '0640',
+    owner        => root,
+    group        => $dns::params::group,
+    mode         => '0640',
+    validate_cmd => "${dns::params::named_checkconf} %",
```
ekohl: Since the `dns` class inherits from `dns::params`, you can use the following which allows making it a parameter as well.

```suggestion
    validate_cmd => "${dns::named_checkconf} %",
```

manifests/config.pp
```diff
@@ -19,13 +20,24 @@
...
     order   => '01',
   }
 
-  concat { [$dns::namedconf_path, $dns::optionspath]:
+  concat { $dns::namedconf_path:
+    owner        => root,
+    group        => $dns::params::group,
+    mode         => '0640',
+    require      => Concat[$dns::optionspath],
+    validate_cmd => "${dns::params::named_checkconf} %",
```
ekohl:
```suggestion
    validate_cmd => "${dns::named_checkconf} %",
```

manifests/config.pp
antaflos: My initial thought was to handle `named_checkconf` in a similar way to `rndcconfgen`, which is not exposed as a class parameter explicitly. But I will implement your suggestion at once :)

antaflos:
Bumping gently :)

ekohl:
Thanks!

PR #145, opened 2019-05-21
Add support for managing sysconfig settings

antaflos:
Debian and RedHat allow setting startup options and other settings via a
sysconfig/default file (/etc/default/bind9 under Debian and
/etc/sysconfig/named under RedHat). FreeBSD/DragonFly and ArchLinux
don't have a direct equivalent.

This change allows managing sysconfig/default settings under Debian and
RedHat. It is especially useful for setting startup options like `-4` to
disable IPv6 support in named, or setting the debug level with `-d`. The
change adds a few specific `$dns::sysconfig_*` parameters to model the
default behaviour under Debian and RedHat, as well as a
`$dns::sysconfig_additional_settings` hash parameter that allows setting
arbitrary key-value pairs in the sysconfig file.

For FreeBSD/DragonFly and ArchLinux these parameters are all set to
undef and not used anywhere since they feature is no directly equivalent
sysconfig setup.

This change also adds basic spec tests for Debian, including tests for
the sysconfig support introduced here.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -59,4 +59,18 @@ class dns::config {
     group => $dns::params::group,
     mode  => '0640',
   }
+
+  # Only Debian and RedHat OS provide a sysconfig or default file where we can
+  # set startup options and other environment settings for named. In FreeBSD
+  # such settings must be set in the global, common /etc/rc.conf file and under
+  # ArchLinux we must use systemd override files to change the startup
+  # commandline. These cases are outside of this module's scope.
+  if $facts['osfamily'] in ['Debian', 'RedHat'] {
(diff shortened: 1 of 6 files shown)


antaflos:
@ekohl It seems that Ubuntu 18.04 (and probably Debians as well) still uses `/etc/init.d/bind9` to manage the service so systemd drop-in files won't help there :(

antaflos:
It seems I was wrong. Both Ubuntu 16.04 and 18.04 provide systemd unit files for bind9, but both read `/etc/default/bind9` as `EnvironmentFile` and pass the `$OPTIONS` variable to the ´ExecStart` commandline, so this change still makes sense.

PR #146, opened 2019-05-29
Allow puppetlabs/concat and puppetlabs/stdlib 6.x

alexjfisher:


metadata.json
```diff
@@ -18,11 +18,11 @@
   "dependencies": [
     {
       "name": "puppetlabs/concat",
-      "version_requirement": ">= 1.0.0 < 6.0.0"
+      "version_requirement": ">= 4.1.0 < 7.0.0"
```
alexjfisher: Minimum version of `1.0.0` seemed fairly nonsensical. `4.1.0` is the first version to list puppet 5 as supported.

PR #149, opened 2019-09-22
Add manage_service parameter

flyingstar16:
This allows users to decide whether the module should manage the
dns service. The check defaults to true, but in case users want to
manage the service autonomously they can.
This also solves "duplicate declaration" issues when the service
is declared in another .pp file

This does not influence package installation.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -18,6 +18,12 @@
 #   Should this module manage the Unix system group under which BIND runs (see
 #   dns::params)?  Defaults to true. Set to false if you want to manage the
 #   system group yourself.
+# @param manage_service
+#   Should this module manage the dns service?
+#   This only applies to the service management (running, stopped) and not to
+#   whether the service should be installed or not.
+#   Calls to "notify => Service['$dns::namedservicename']" will still be
+#   executed
 # @param namedservicename
 #   Name of the service
 # @param zonefilepath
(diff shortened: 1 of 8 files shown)


manifests/init.pp
```diff
@@ -18,6 +18,12 @@
 #   Should this module manage the Unix system group under which BIND runs (see
 #   dns::params)?  Defaults to true. Set to false if you want to manage the
 #   system group yourself.
+# @param manage_service
+#   Should this module manage the dns service?
+#   This only applies to the service management (running, stopped) and not to
+#   whether the service should be installed or not.
+#   Calls to "notify => Service['$dns::namedservicename']" will still be
```
ekohl: That's actually going to be a problem I think. Please add tests for that. I suspect we need to change those to `Class['dns::service']` instead.

flyingstar16: In my mind, the calls should always happen because otherwise configuration changes will not be picked up by the service until a manual restart is issued.
Do you think we should inhibit them instead (I'm 99% sure that's what calling `Class['dns::service']` would do, but please correct me if I'm wrong)?

Also, tests. You're right. Forgot about them.
I need to learn Ruby unit testing, might take me a little bit.

ekohl: Correct. If you notify `Class['dns::service']` it'd work. It respects the `manage_service` parameter and knows the correct service name.

If you need help with the tests, let me know. I'm planning a release next week so now's the ideal time to get your changes in :)

flyingstar16:
@ekohl Ok, I've pointed all the calls to Class['dns::service'] and added tests to dns_init_spec.rb for the manage_service parameter.
Can you have a look and let me know if those tests are enough?
I tried adding tests in dns_zone_spec.rb but rspec correctly informed me that you can't test a parameter that the class doesn't have...

spec/defines/dns_zone_spec.rb
```diff
@@ -38,7 +38,7 @@
       :group    => 'named',
       :mode     => '0644',
       :replace  => 'false',
-      :notify   => 'Service[named]',
+      :notify   => 'Class[Dns::Service]',
```
ekohl: A more correct way of checking uses the catalog. I realize this was already incorrect, but I'd appreciate it if you fixed it while changing this part
```suggestion
      ).that_notifies('Class[Dns::Service]')
```

flyingstar16:
Sorry about the horrible delay on a one-line fix :/ Should be ok now! :)

mmoll:
merged, thanks @flyingstar16!

PR #151, opened 2019-11-23
Refactor soaip in dns::zone

ekohl:
This makes the $soaip parameter optional since in many cases it doesn't makes sense. It was also breaking hard on IPv6-only systems. For those setups the $soaipv6 parameter is also added. The types are also tightened to the correct address family.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -1,4 +1,14 @@
-# Define new zone for the dns
+# @summary Define new zone for the dns
+#
+# @param soaip
+#   The IP address for the SOA. If `reverse` is false, an A record will be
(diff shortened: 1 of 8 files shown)


ekohl:
So I can't reproduce the failure on a vagrant_libvirt VM. Also docker running on CentOS7. However, I can on docker-ce 19.03.5.

https://github.com/moby/moby/issues/38749 / https://github.com/docker/for-linux/issues/835

ekohl:
https://github.com/theforeman/puppet-dns/pull/152 to work around this.

mmoll:
please rebase :)

ekohl:
:green_heart:

mmoll:
merged, bedankt @ekohl!

PR #152, opened 2019-11-23
Test on CentOS 6 instead of CentOS 7

ekohl:
Due to a bug in Docker CE named can't be started because it messes up
the cgroups.


mmoll:
let's make this :green_heart: again!

PR #154, opened 2020-02-05
Fixes #28939 - Replace DNS zone file

jeremylenz:
When running `satellite-change-hostname`, it was not updating DNS resource records when applicable.

For the DNS zone file, this update sets the `replace` value to `true` instead of false
* Replaces DNS zone files in `/var/named/dynamic` with the correct contents after hostname change
* Only replaces the file if its `md5sum` is different
* Will happen automatically since `satellite-change-hostname` runs `foreman-installer`.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -97,7 +97,7 @@ define dns::zone (
       group   => $dns::group,
       mode    => '0644',
       content => template('dns/zone.header.erb'),
-      replace => false,
+      replace => true,
       notify  => Class['dns::service'],
     }
   }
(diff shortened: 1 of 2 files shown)


chris1984:
Starting to test

jeremylenz:
On a `sat-deploy` box, you can test this fix by editing `/usr/share/foreman-installer/modules/dns/manifests/zone.pp`

jeremylenz:
https://projects.theforeman.org/issues/28939

ekohl:
You can't do this because you'll destroy user data. The zone file contains all user modifications and systems they might have provisioned. The effect is that every time the module is applied, you wipe the database. This needs to be solved in the hostname change script. It should probably call APIs on the smart-proxy.

Issue #156, opened 2020-03-08
Missing zone options


damluk:
I would like to set several options inside zones that are currenty not supported by `named.zone.erb`. Would you like to make me a merge request with a bunch of new `dns::zone` parameters or with an `additional_options` parameter?

ekohl:
A PR with that would be welcome. It depends on the number of options and how well defined they are, but I think a bunch of parameters is where I'd start.

damluk:
A zone definition supports ~75 different parameters. Do I need to be exhaustive in the PR? I am currently only interested in about five of them.

ekohl:
No need for an exhaustive list, just the ones you care about. It's how I always do PRs :)

ekohl:
With both PRs merged, is there still something to do here?

damluk:
> With both PRs merged, is there still something to do here?

Nope, we are done, thanks. :)

PR #157, opened 2020-03-08
Make zone update_policy_rules more generic

damluk:
This is part of https://github.com/theforeman/puppet-dns/issues/156. I submit it separately from the other zone options I want to introduce, because this PR might be a breaking change for some installations.


damluk:
I decided to deprecate the parameter instead of breaking directly.

PR #160, opened 2020-03-30
Fixes #29210 - support el8

wbclark:


wbclark:
@ekohl I noticed el7 tests were missing here so I added those as well. If that's intentional, please let me know why and I'll remove them. Thanks!

ekohl:
See https://github.com/theforeman/puppet-dns/commit/bf846e44b901eb51b1e15e660e339a11b5cabcce / https://github.com/theforeman/puppet-dns/pull/151#issuecomment-557836499 for the EL7 history. It may work to use the CentOS 7.6 container image instead of the latest 7.

mmoll:
@wbclark Also, needs a rebase.

.sync.yml
```diff
@@ -5,6 +5,8 @@
 .travis.yml:
   beaker_sets:
     - centos6-64
+    - centos7-64
```
ekohl, line 8: This will fail: https://github.com/theforeman/foreman-installer-modulesync/pull/121 should have the workaround.

spec/setup_acceptance_node.pp
```diff
@@ -0,0 +1,8 @@
+if $facts['os']['family'] == 'RedHat' {
```
ekohl, line 1: As stated elsewhere: I'd like to avoid this if possible.

wbclark: Will update this once I see how https://github.com/theforeman/puppet-dhcp/pull/167/files#diff-354f30a63fb0907d4ad57269548329e3R38 works out

wbclark:
Due to https://github.com/docker/for-linux/issues/835 , services are failing to start in the Travis test environment, causing test failures.

However all tests are succeeding when I run them locally on a vagrant_libvirt el8 VM.

@ehelms what are your thoughts on adding el8 support without adding tests to the travis config, similarly to what we previously did for this project on el7?

ehelms:
I think this sounds like a good approach given the limiting factors of our testing environment. I'll defer to @ekohl or @mmoll as primary maintainers of the module.

mmoll:
I am OK with this, as a fix for the time after 2.1 branching seems to be in the works.

mmoll:
merged, thanks @wbclark!

PR #161, opened 2020-04-11
Modulesync: use voxpupuli helper gems

ekohl:


mmoll:
merged, bedankt @ekohl!

PR #162, opened 2020-04-16
Modulesync: require voxpupuli-test 1.2

ekohl:
Follow up on 01d1ede67a89dd3d8e0a2f4edcd9f2de3971735a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ junit/
 pkg/
 coverage/
 .yardoc/
+REFERENCE.md
 
 ## InteliJ / RubyMine
 .idea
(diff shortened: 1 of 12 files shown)


ekohl:
@wbclark why? I don't see any overrides we need.

wbclark:
> @wbclark why? I don't see any overrides we need.

I think I will need some for EL8 but can add the file when that becomes necessary.

By the way, looking at the failed spec tests, it looks to me like facts are not loaded properly. Any idea why?

ekohl:
I spent quite a bit of time to convert testing to rspec-puppet-facts but it's not complete. I did want to push my work though.

ekohl:
This patch turned out much larger than anticipated due to the conversion to rspec-puppet-facts.

ekohl:
:green_heart:

mmoll:
merged, bedankt @ekohl!

PR #163, opened 2020-05-01
Make bind view clauses optional

dlucredativ:
bind 9.16 deprecated the `dnssec-enable` clause. Use this opportunity to make all output clauses optional.


ekohl:
Thanks! Since it's changing defaults I've marked it as Backwards Incompatible, thought the changes required should be quite minimal.

PR #172, opened 2020-07-02
Add the ability to define logging

coreone:
There was no way in the current settings to configure a log directory or an advanced logging configuration using channels and categories.  This change adds the necessary configuration options and template changes to create a full-featured logging configuration.  README and spec changes have been completed to document and test the change.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -24,7 +24,7 @@ class dns::config {
   }
 
   concat { $dns::namedconf_path:
-    owner        => root,
+    owner        => 'root',
     group        => $dns::params::group,
     mode         => '0640',
     require      => Concat[$dns::optionspath],
@@ -34,7 +34,7 @@ class dns::config {
   # This file cannot be checked by named-checkconf because its content is only
   # valid inside an "options { };" directive.
   concat { $dns::optionspath:
-    owner => root,
+    owner => 'root',
     group => $dns::params::group,
     mode  => '0640',
   }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -14,6 +14,8 @@
 #   Path of the config file holding all the zones
 # @param vardir
 #   Directory holding the variable or working files
+# @param logdir
+#   Directory holding the log files for named
 # @param group_manage
 #   Should this module manage the Unix system group under which BIND runs (see
 #   dns::params)?  Defaults to true. Set to false if you want to manage the
@@ -119,9 +121,15 @@
 #   A hash of zones to be created. See dns::zone for options.
 # @param keys
 #   A hash of keys to be created. See dns::key for options.
+# @param logging_categories
+#   A hash of logging categories to be created. See dns::logging::category for options.
+# @param logging_channels
+#   A hash of logging channels to be created. See dns::logging::channel for options.
 #
 # @see dns::zone
 # @see dns::key
+# @see dns::logging::category
+# @see dns::logging::channel
 class dns (
   Stdlib::Absolutepath $namedconf_path                              = $dns::params::namedconf_path,
   Stdlib::Absolutepath $dnsdir                                      = $dns::params::dnsdir,
@@ -130,6 +138,7 @@ class dns (
   Stdlib::Absolutepath $optionspath                                 = $dns::params::optionspath,
   Stdlib::Absolutepath $publicviewpath                              = $dns::params::publicviewpath,
   Stdlib::Absolutepath $vardir                                      = $dns::params::vardir,
+  Stdlib::Absolutepath $logdir                                      = $dns::params::logdir,
   Boolean $group_manage                                             = $dns::params::group_manage,
   Boolean $manage_service                                           = $dns::params::manage_service,
   String $namedservicename                                          = $dns::params::namedservicename,
@@ -164,6 +173,8 @@ class dns (
   Boolean $enable_views                                             = $dns::params::enable_views,
   Hash[String, Hash] $zones                                         = $dns::params::zones,
   Hash[String, Hash] $keys                                          = $dns::params::keys,
+  Hash[String, Hash] $logging_categories                            = $dns::params::logging_categories,
+  Hash[String, Hash] $logging_channels                              = $dns::params::logging_channels,
 ) inherits dns::params {
 
   include dns::install
@@ -174,4 +185,6 @@ class dns (
 
   create_resources('dns::key', $keys)
   create_resources('dns::zone', $zones)
+  create_resources('dns::logging::category', $logging_categories)
+  create_resources('dns::logging::channel', $logging_channels)
 }
--- a/manifests/logging/channel.pp
+++ b/manifests/logging/channel.pp
@@ -1,0 +1,71 @@
+# Define new channel for logging
+#
+# @param file_path
+#   The path to the log file
+#
+# @param file_size
+#   The maximum size the log file is allowed to reach
+#
+# @param file_versions
+#   The number of log files to keep when rotating
+#
+# @param log_type
+#   The destination type for the log (file, stderr, syslog, or "null")
+#
+# @param order
+#   The order of the channel in the configuration file
+#
+# @param print_category
+#   Decide whether to log the category in the log message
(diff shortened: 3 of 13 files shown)


manifests/init.pp
```diff
@@ -164,6 +165,7 @@
   Boolean $enable_views                                             = $dns::params::enable_views,
   Hash[String, Hash] $zones                                         = $dns::params::zones,
   Hash[String, Hash] $keys                                          = $dns::params::keys,
+  Optional[Hash] $logging                                           = {},
```
ekohl, line 168: Would it make sense to add a a type for this to document/verify the passed in datastructure?

coreone: I converted this to being implemented as 2 different defined types.  If it looks reasonable, I will squash/rebase this branch.

coreone: @ekohl This was a better solution, so I squashed and rebased.

coreone:
Is there anything else needed here?

manifests/logging/channel.pp
```diff
@@ -0,0 +1,43 @@
...
+  Integer $order                                     = 10,
+  Optional[Enum['no', 'yes']] $print_category        = undef,
+  Optional[Enum['no', 'yes']] $print_severity        = undef,
+  Optional[Enum['no', 'yes']] $print_time            = undef,
+  Optional[String] $severity                         = undef,
+  Optional[String] $syslog_facility                  = undef,
+) {
+  $channel_name = $title
+
+  if ($order < 2) or ($order > 98) {
```
ekohl, line 16: It'd recommend using `Integer[2, 98]` for `$order` so you don't need to write this check.

spec/classes/dns_init_spec.rb
```diff
@@ -116,6 +118,7 @@
           end
 
           verify_concat_fragment_exact_contents(catalogue, 'options.conf+10-main.dns', expected)
+
```
ekohl, line 121: Nit: could you drop this line?

spec/classes/dns_init_spec.rb
```diff
@@ -132,14 +135,19 @@
             '};',
             "include \"#{localzonepath}\";",
             '// Public view read by Server Admin',
-            "include \"#{etc_named_directory}/zones.conf\";"
+            "include \"#{etc_named_directory}/zones.conf\";",
+            "// Include logging settings",
+            "include \"#{etc_named_directory}/logging.conf\";"
```
ekohl, line 140: Feel free to add a trailing comma here

manifests/config.pp
```diff
@@ -55,6 +71,18 @@
     order   => '10',
   }
 
+  concat::fragment { 'logging.conf+1-header.dns':
+    target  => $dns::logconfpath,
+    content => template('dns/logging/log.header.conf.erb'),
```
ekohl, line 76: Since these are so trivial, you can inline this (and the footer):
```suggestion
    content => "logging {\n",
```

spec/acceptance/logging_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      it { is_expected.to be_listening }
+    end
+
+    describe command('dig +short SOA example.com @localhost') do
+      its(:stdout) { is_expected.to match("ns1.example.com. root.example.com. 1 86400 3600 604800 3600\n") }
+    end
+
+    describe command('cat /var/log/named/queries.log') do
+      its(:stdout) { is_expected.to match(%r{query: example\.com IN SOA \+}) }
+    end
```
ekohl, line 52: I'd recommend to use the [file resource](https://serverspec.org/resource_types.html#file).

coreone: Done.

manifests/config.pp
```diff
@@ -43,6 +50,15 @@
     mode  => '0640',
   }
 
+  concat { $dns::logconfpath:
```
ekohl, line 53: What is the reason you need a separate log file for this instead of adding it to the main concat?

templates/log.channel.conf.erb
```diff
@@ -0,0 +1,26 @@
+  channel <%= @channel_name %> {
+    <%- if @log_type == 'file' -%>
+    file "<%= @file_path -%>" versions <%= @file_versions -%> size <%= @file_size -%>;
+    <%- end -%>
+    <%- if @log_type == 'null' -%>
```
ekohl, line 5:
```suggestion
    <%- elsif @log_type == 'null' -%>
```

spec/defines/dns_logging_channel_spec.rb
```diff
@@ -0,0 +1,248 @@
...
+
+      context 'fail when order => 1' do
+        let(:params) do
+          {
+            :log_type => 'stderr',
+            :order    => 1,
+          }
+        end
+
+        it { is_expected.to_not compile }
```
ekohl, line 232: You can also use this to check that it compiles because order failed it:
```suggestion
        it { is_expected.to compile.and_raise_error(%r{order}) }
```

(same below for order 99)

templates/log.channel.conf.erb
```diff
@@ -0,0 +1,26 @@
...
+    null;
+    <%- end -%>
+    <%- if @log_type == 'stderr' -%>
```
ekohl, line 8:
```suggestion
    <%- elsif @log_type == 'stderr' -%>
```

templates/log.channel.conf.erb
```diff
@@ -0,0 +1,26 @@
...
+    stderr;
+    <%- end -%>
+    <%- if @log_type == 'syslog' -%>
```
ekohl, line 11:
```suggestion
    <%- elsif @log_type == 'syslog' -%>
```

manifests/config.pp
```diff
@@ -12,6 +12,13 @@
     validate_cmd => "${dns::named_checkconf} %",
   }
 
+  file { $dns::logdir:
+    ensure => directory,
+    owner  => $dns::params::user,
+    group  => $dns::params::group,
+    mode   => '0755',
+  }
```
ekohl, line 20: This is only needed if there is a log channel with type file, right? Can we include it only if needed to keep the number of managed files down?

spec/classes/dns_init_spec.rb
```diff
@@ -151,6 +158,8 @@
         it { should contain_file(rndc_key) }
 
         it { should contain_service(service_name).with_ensure('running').with_enable(true).with_restart(nil) }
+
+        it { should contain_file("#{logdir}").with_ensure('directory') }
```
ekohl, line 162: No need to interpolate here
```suggestion
        it { should contain_file(logdir).with_ensure('directory') }
```

spec/classes/dns_init_spec.rb
```diff
@@ -169,7 +178,9 @@
           "        include \"#{options_path}\";",
           '};',
           '// Public view read by Server Admin',
-          "include \"#{etc_named_directory}/zones.conf\";"
+          "include \"#{etc_named_directory}/zones.conf\";",
+          "// Include logging settings",
+          "include \"#{etc_named_directory}/logging.conf\";"
```
ekohl, line 183: Feel free to add a trailing comma
```suggestion
          "include \"#{etc_named_directory}/logging.conf\";",
```

manifests/config.pp
coreone: This was primarily due to the number of things that can be logged with named.  If I turn on all the logging options, that file is 154 lines long.  I thought potentially putting all of that in the main config file would be too much.

manifests/config.pp
coreone: I'm not sure how I would selectively ensure this directory if and only if a defined type was instantiated with specific options.

coreone:
@ekohl I refactored this to append to `named.conf` instead of creating a new config file.  I believe that was the last piece for this PR.

coreone:
@ekohl I think this should be all set now.  Let me know if there is anything else you would want updated before merging.

ekohl (approved):
Thanks!

PR #173, opened 2020-09-03
Fix custom key ordering

coreone:
When defining a custom key, the configuration validation would fail on the first run as the configuration would be updated and the validation would run before the key was created.  The error would go away on the second run as the file would then exist before the configuration was updated (from the previous run).  The updated ordering in `dns::key` makes this now work on the first run without an error.

As a note, the `group` creation needed to be moved out of `dns::config` to prevent a dependency cycle.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -3,5 +3,14 @@
 class dns::install {
   if ! empty($dns::dns_server_package) {
     ensure_packages([$dns::dns_server_package])
+    $pkg_req = Package[$dns::dns_server_package]
+  } else {
+    $pkg_req = undef
+  }
+
+  if $dns::group_manage {
+    group { $dns::group:
+      require => $pkg_req,
+    }
   }
 }
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -1,10 +1,6 @@
 # Configure dns
 # @api private
 class dns::config {
-  if $dns::group_manage {
-    group { $dns::params::group: }
-  }
-
   concat { $dns::publicviewpath:
     owner        => root,
     group        => $dns::params::group,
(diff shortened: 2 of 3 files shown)


manifests/init.pp
```diff
@@ -170,6 +170,10 @@
   include dns::config
   contain dns::service
 
+  if $group_manage {
+    group { $group: }
+  }
```
ekohl, line 175: Normally the group is created by the package in install.pp. Perhaps it should be there and also chained to first install the package and then manage the group.

manifests/install.pp
```diff
@@ -3,5 +3,14 @@
 class dns::install {
   if ! empty($dns::dns_server_package) {
     ensure_packages([$dns::dns_server_package])
+    $pkg_req = Package[$dns::dns_server_package]
+  } else {
+    $pkg_req = undef
+  }
+
+  if $dns::group_manage {
+    group { $dns::params::group:
```
ekohl, line 12: This was already incorrect previously, but mind fixing it while you're at it?
```suggestion
    group { $dns::group:
```

manifests/init.pp
coreone: Moved to `install.pp` with dependencies.

manifests/install.pp
coreone: Fixed

ekohl (commented):
At first glance this looks good. Let's see what the tests show.

coreone:
@ekohl is there anything else needed here?

ekohl (approved):
Thanks for the ping and contribution. I notice I often forget to check back on a PR when I need to wait for CI. Merging.

PR #174, opened 2020-10-12
Update zone documentation

marcdeop:
`update_policy_rules` was removed some time ago but I still found a reference to its usage in the zone.pp puppet-strings documentation.

This PR updates the docs parameter as well as adding all the other missing ones (even though it doesn't define them, it's better than nothing ;-) ).

Thanks to [Erasys GmbH](https://erasys.de) for their support

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -16,7 +16,7 @@
 # @param manage_file_name
 #   Whether to set the file parameter in the zone file.
 #
-# @param update_policy_rules
+# @param update_policy
 #   This can be used to specifiy additional update policy rules in the
 #   following format
 #   { '<KEY_NAME' => {'matchtype' => '<VALUE>', 'tname' => '<VALUE>', 'rr' => 'VALUE' } }
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -23,6 +23,32 @@
 #   Example {'foreman_key' => {'matchtype' => 'zonesub', 'rr' => 'ANY'}}
 #   tname and rr are optional
 #
+# @param target_views
+# @param zonetype
+# @param soa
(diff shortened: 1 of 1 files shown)


manifests/zone.pp
```diff
@@ -16,13 +16,39 @@
...
+# @param contact Optional[String]
+# @param zonefilepath Stdlib::Absolutepath
+# @param filename String
+# @param forward Enum['first', 'only']
+# @param forwarders Array
+# @param dns_notify Optional[Enum['yes', 'no', 'explicit']]
+# @param key_directory Optional[Stdlib::Absolutepath]
+# @param inline_signing Optional[Enum['yes', 'no']]
+# @param dnssec_secure_to_insecure Optional[Enum['yes', 'no']]
+# @param auto_dnssec Optional[Enum['allow', 'maintain', 'off']]
```
ekohl, line 50: I think append the type this way doesn't really work. puppet-strings infers the type if there is one. That means you can leave it off. Ideally we would actually have documentation what they do.

ekohl (commented):
Sadly the lint check that would catch this is disabled:
https://github.com/theforeman/puppet-dns/blob/820cb719907aac3eb7e28a506eac72a236f846e3/.puppet-lint.rc#L4

If we would have all parameters documented, it could actually catch this exact problem.

manifests/zone.pp
marcdeop: You are right, the type doesn't need to be there.

I haven't added what the params do because I don't really know. My intention was only to fix the mistake with `update_policy_rules`. Adding the params is only an extra here. You can look at it as a "baby step" :-)

ekohl (approved):

PR #175, opened 2020-10-13
Replace get_in_addr_arpa with reverse_dns function

ekohl:
The get_in_addr_arpa function did not support IPv6. Its implementation was also the old legacy function API. This replacement also has tests, data types and documentation.


ehelms (approved):
@ekohl OK to merge this and include in the upcoming release?

ekohl:
Yes

Issue #164, opened 2020-05-01
ACLs are unusable with allow_transfer


damluk:
The following beaker test fails on Debian 10 and probably other distros, even though it might encode a desirable setup:
```
diff --git a/spec/acceptance/dns_spec.rb b/spec/acceptance/dns_spec.rb
index 50f6729..8ddc748 100644
--- a/spec/acceptance/dns_spec.rb
+++ b/spec/acceptance/dns_spec.rb
@@ -14,12 +14,15 @@ describe 'Scenario: install bind' do
 
   let(:pp) do
     <<-EOS
-    include dns
+    class { 'dns':
+      acls => { 'trusted' => ['10.0.0.0/24'] },
+    }
 
     dns::zone { 'example.com':
       soa     => 'ns1.example.com',
       soaip   => '192.0.2.1',
       soaipv6 => '2001:db8::1',
+      allow_transfer => ['"trusted"'],
     }
     EOS
   end

```
The error is due to named-checkconf:
```
Error: Execution of '/usr/sbin/named-checkconf /etc/bind/zones.conf20200501-1763-1xd9g1x' returned 1: /etc/bind/zones.conf20200501-1763-1xd9g1x:7: undefined ACL 'trusted'
```
Interestingly there are other ACL-capable clauses that do not care about undefinedness in the same config file.

The best solution I can come up with, is to place acl, view and zone definitions into the same config file.

ekohl:
Looks like this problem is not limited to Debian 10.

ikonia:
bumping this it looks like dns::acl doesn't actually define the ACL into any config file, meaning any parameter that depends on an ACL will fail, I've just hit this tightening my query options, I define an ACL with dns::acl yet the acl appears in no config file so bind 9.18 refuses to start due to the ACL being undefined

PR #178, opened 2020-10-29
Allow configuration checks to be turned off

coreone:
I was having the same issue as described in #164 so I added this so the checks could be turned off if needed.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -109,6 +109,10 @@
 #   setting `service_restart_command` to `/usr/sbin/service bind9 reload` or
 #   `/usr/sbin/rndc reload` or even `/usr/bin/systemctl try-reload-or-restart bind9`.
 #   Default is 'undef' so the service resource default is used.
+# @param config_check
+#   Should this module run configuration checks before putting new configurations in
+#   place?  Defaults to true. Set to false if you don't want configuration checks when
+#   config files are changed.
 # @param additional_options
 #   Additional options
 # @param additional_directives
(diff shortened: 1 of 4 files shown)


spec/classes/dns_init_spec.rb
```diff
@@ -365,6 +365,21 @@
         it { should contain_file("#{etc_directory}/dns-key.key") }
       end
 
+      describe 'with config_check set to false' do
+        let :params do
```
alexjfisher, line 369: For consistency with the other tests, maybe
`let(:params) { {:config_check => false} }`??

spec/classes/dns_init_spec.rb
```diff
@@ -365,6 +365,21 @@
...
+          {
+            'config_check' => false,
+          }
+        end
+
+        it { is_expected.to compile.with_all_deps }
+
+        it {
+          is_expected.to contain_concat("#{etc_directory}/named.conf")
+            .with_validate_cmd(nil)
```
alexjfisher, line 379: Consider `.without_validate_cmd`

alexjfisher (approved):
LGTM other than a couple of very minor style suggestions.

manifests/init.pp
```diff
@@ -168,6 +168,7 @@
   Variant[Enum['running', 'stopped'], Boolean] $service_ensure      = $dns::params::service_ensure,
   Boolean $service_enable                                           = $dns::params::service_enable,
   Optional[String[1]] $service_restart_command                      = $dns::params::service_restart_command,
+  Boolean $config_check                                             = $dns::params::config_check,
```
alexjfisher, line 171: You need to add a puppet strings style documentation comment.

alexjfisher (changes requested):

alexjfisher (approved):

ekohl (approved):

PR #179, opened 2020-10-29
Switch to GH Actions from Travis

wbclark:


wbclark:
Removed centos7 and centos8 from test matrix. These were already not running Travis.

ekohl:
I opened https://github.com/theforeman/puppet-dhcp/pull/180 which includes all changes from https://github.com/theforeman/foreman-installer-modulesync/pull/106. Also has a workaround for EL7 testing.

PR #181, opened 2020-11-04
Drop Fedora 26, add Fedora 32

ekohl:


spec/setup_acceptance_node.pp
```diff
@@ -7,3 +7,10 @@
 package { $package:
   ensure => installed,
 }
+
+# Needed for the ss command in testing and not included in the base container
```
ekohl, line 11: This would need a fix in beaker-hostgenerator but reviews there are slow so here is the workaround.

ehelms (approved):

PR #182, opened 2020-12-22
Test IPv6 in docker: Lets see what CI says

b4ldr:


b4ldr:
apparently fedora dosn;t have sysctl which is a surprise and i think the cantos 6 faliures are related to that being EOL, but works on the others.  could always just `echo 1 > /proc/sys/net/ipv6/conf/all/disable_ipv6`.  previously i used sysctl as i needed to support freebsd which dosn't mount `/proc/sys`

ekohl:
I think the Fedora cloud image is very minimal. I wonder what happens if you install it prior to running the command.
And you're right about CentOS 6. I need to clean up some code.

b4ldr:
> I think the Fedora cloud image is very minimal. I wonder what happens if you install it prior to running the command.
And you're right about CentOS 6. I need to clean up some code.

Not sure i just tested localy and the default image i pull down has `/proc/sys` mounted ro as well.  I have just sent an update to see if using echo works better

b4ldr:
ok that works better, ill update the other PR, feel free to close this as you would, im also jbond42 on irc if you want to ping me there

ekohl:
Thanks for checking that out. I think we've learned enough about it for now. Applied it here instead:
https://github.com/theforeman/puppet-dns/blob/2f112796cf8d55b1783e76d87c2a890294e8cf7f/spec/setup_acceptance_node.pp#L18-L21

PR #191, opened 2021-06-01
Drop Puppet 5 support

ehelms:


karelyatin:
Would be good to drop CentOS 6 testing, it's EOL and also repos no longer available for it http://mirror.centos.org/centos/6/readme.

Also debian 8 failing as http://apt.puppet.com/puppet7-release-jessie.deb no longer exists, likely this can also be removed from metadata.json

Noticed these issues in my PR https://github.com/theforeman/puppet-dns/pull/190/checks, better to do cleanup in this patch itself.

ehelms:
To not fix concerns, I opened https://github.com/theforeman/puppet-dns/pull/193 for OS support changes

ekohl (approved):

PR #201, opened 2022-01-27
Support CentOS 9, Debian 11 and Ubuntu 20.04

ekohl:


wbclark (approved):

wbclark:
not a new issue with this PR but it looks like centos8 doesn't get included in the test matrix due to a necessary workaround implemented in puppet_metadata... is it possible to test with centos8-stream without marking it as a supported OS in metadata?

ekohl:
> not a new issue with this PR but it looks like centos8 doesn't get included in the test matrix due to a necessary workaround implemented in puppet_metadata... is it possible to test with centos8-stream without marking it as a supported OS in metadata?

It should be supported since it's already in metadata. The reason it's now skipped is that in the past we had problems with the `PIDFile=` part in systemd which broke under Docker. I've opened https://github.com/theforeman/puppet-dns/pull/202 to see if that's still really a problem.

PR #202, opened 2022-01-31
Drop PIDFile workaround

ekohl:
This should test it on CentOS 8 as well.


ekohl:
@wbclark as you can see, this is why we don't test on CentOS 8.

PR #206, opened 2022-02-16
Add support for defining custom dnssec-policies

smortex:
This is a complement to #205 to allow to setup custom policies.

This is currently a draft until I manage to deploy this code in a real-world workload.

### DIFF
--- a/spec/defines/dns_dnssec_policy_spec.rb
+++ b/spec/defines/dns_dnssec_policy_spec.rb
@@ -1,0 +1,92 @@
+require 'spec_helper'
+
+describe 'dns::dnssec_policy' do
+  on_supported_os.each do |os, os_facts|
+    context "on #{os}" do
+      let(:facts) { os_facts }
+      let(:title) { "standard" }
+
+      let :pre_condition do
+        'include dns'
+      end
+
+      context "without params" do
+        it { is_expected.to compile }
+
+        it "should have valid zone file contents" do
+          verify_concat_fragment_exact_contents(catalogue, 'dnssec-policy-standard', [
+            'dnssec-policy standard {',
+            '};',
+          ])
+        end
+
+        context "when using a reserved name" do
+          let(:title) { "default" }
+
+          it { is_expected.to compile.and_raise_error(/The name "default" is reserved and cannot be used/) }
+        end
+      end
+
+
+      context "with all params set" do
+        let(:params) do
+          {
+            dnskey_ttl: 600,
+            keys: [
+              {
+                type: "ksk",
+                directory: "key-directory",
+                lifetime: "unlimited",
+                algorithm: "rsasha1",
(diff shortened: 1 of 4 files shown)


spec/defines/dns_dnssec_policy_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+          verify_concat_fragment_exact_contents(catalogue, 'dnssec-policy-standard', [
+            'dnssec-policy standard {',
+            '};',
+          ])
+        end
+
+        context "when using a reserved name" do
+          let(:title) { "default" }
+
+          it { is_expected.not_to compile }
```
ekohl, line 26: I prefer to check the error message too:
```suggestion
          it { is_expected.to compile.and_raise_error(/is reserved/) }
```

templates/named.dnssec_policy.epp
```diff
@@ -0,0 +1,39 @@
...
+  <%- } -%>
+  <%- if $signatures_validity { -%>
+    signatures-validity <%= $signatures_validity %>;
+  <%- } -%>
+  <%- if $signatures_validity_dnskey { -%>
+    signatures-validity-dnskey <%= $signatures_validity_dnskey %>;
+  <%- } -%>
+  <%- if $zone_propagation_delay { -%>
+    zone-propagation-delay <%= $zone_propagation_delay %>;
+  <%- } -%>
```
ekohl, line 38: I wonder if it would make sense to pass this as a hash to `epp()` so you don't need so many `if` statements. Either way is fine though.

ekohl (commented):
Some very minor suggestions, overall it looks good to me.

templates/named.dnssec_policy.epp
smortex: No strong feeling here… This quite looks like PHP but I can live with this and doing the "formatting" on the Puppet side seems less appropriate I think.

ekohl: Perhaps. You could use `$context = { 'keys' => ..., 'options' => { .... } }` but I see your point. Either way works for me.

smortex: I just pushed a commit to confirm I understood what you say… It feels worse AFAIAC :sob: …

smortex: Let me know what you prefer; I'll either squash or drop the second commit ;-)

ekohl (commented):
I'm still not entirely sure.

ekohl (approved):
It looks good to me, but perhaps @evgeni or @wbclark can weigh in on their preference.

evgeni:
I have no strong opinion (and not too versed in dnssec), but it doesn't look crazy to me? ;)

smortex:
Thanks!  I squashed the commits.

Issue #196, opened 2021-07-23
Manage zone records


BDelacour:
Hello,

Is there a way to manage zones records ?
I would like Puppet to enforce my complete bind configuration.

Actually, `zone.header.erb` only manages the header of the file.
Is there a way without concat to add my records (A, CNAME, ...) at the bottom ?

Thank you !

ekohl:
Today it's not easy. You can set `manage_file` to false and manually manage it.

This is mostly because in our use case we allow updates via `nsupdate`. Since it's all dynamic at runtime, the file must not be overwritten. This is a slightly different model than where Puppet is fully in control.

rtprio:
Is this a feature that you'd be receptive to merging, even if it's a feature you wouldn't be using?

Other modules do this, but don't have such broad support (like FreeBSD).

ekohl:
I would be. As long as it's clear and works consistently. Ideally speaking it would even disallow nsupdate commands to zones that are managed by Puppet to avoid nasty situations where Puppet reverts changes, but I'd be happy with a clear note (either in readme or in class/definition documentation that you should be aware of this.

rtprio:
Seems more cluttered when I try to do it this way, and it changes the zone file resource into a concat resource. Which I guess isn't a problem but complicates both the code and the tests.

PR #212, opened 2022-05-03
Manage zone records

BDelacour:
See #196 

@ekohl Is it OK for you to manage zone without validation or would you prefer a more specific type (which would lead to a configuration more difficult to maintain) ?

Example :
- I created something like this (in Hiera)
```
dns::zones:
  example.com:
    replace_file: true
    serial: 2
    records:
      - host1 IN A 192.168.0.1
      - host2 IN A 192.168.0.2
```
- Or you want something like this (don't mind the naming, just for the format)
```
dns::zones:
  example.com:
    replace_file: true
    serial: 2
    records:
      - name: host1
         xx: IN
         type: A
         dst: 192.168.0.1
...
```

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -37,6 +37,10 @@
 # @param negttl
 # @param serial
 # @param records
+#   A list of records which will be added to the zone file in
+#   the RFC 1035 format (see https://datatracker.ietf.org/doc/html/rfc1035)
+#   Example ['host1 IN A 192.168.0.10', 'alt-host1 IN CNAME host1']
+#
 # @param masters
 # @param allow_transfer
 # @param allow_query
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -16,6 +16,9 @@
 # @param manage_file_name
 #   Whether to set the file parameter in the zone file.
 #
+# @param replace_file
+#   Whether to update the zone file when a change is detected.
+#
 # @param update_policy
 #   This can be used to specifiy additional update policy rules in the
 #   following format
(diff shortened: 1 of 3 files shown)


manifests/zone.pp
```diff
@@ -64,6 +68,7 @@
   Integer $expire                                         = 604800,
   Integer $negttl                                         = 3600,
   Integer $serial                                         = 1,
+  Array $records                                          = [],
```
ekohl, line 71: Perhaps tighten this? to only allow non-empty strings?
```suggestion
  Array[String[1]] $records                               = [],
```

manifests/zone.pp
```diff
@@ -33,6 +36,7 @@
 # @param expire
 # @param negttl
 # @param serial
+# @param records
```
ekohl, line 39: Could you add a description here? We haven't done a great job in the past, but by documenting at least what we add we keep it in reasonable shape.

ekohl (approved):
Since you explicitly pinged me, I gave it a quick look but I'm on leave so officially not working. @evgeni / @ehelms would you mind taking care of this? I just approved CI so the tests can run and some minor inline comments on small details, but the general idea looks good to me.

manifests/zone.pp
BDelacour: Sorry for bothering you with this PR, I pinged you because you commented in the issue ^^.

I added a description in 487857bc6349ffc64d8bd9cced429a1f06c13942

evgeni (approved):

PR #217, opened 2022-09-07
Added defaultzonepath to template name.conf.erb

### LINKED ISSUE #216: defaultzonepath parameter defined, but not used in template name.conf.erb
dns::defaultzonepath is defined as a parameter, however setting this has no effect to named.conf.

benjamin-robertson:
Fixes issue #216 

Closes #216


benjamin-robertson:
Closes #216

ekohl (approved):
I think this is correct. Prior to this it was only used when views were enabled (https://github.com/theforeman/puppet-dns/blob/47fb22b2812302167b0003bc1479346ffa3b199c/templates/named.view_header.erb#L50-L54) but this is in a section where views aren't enabled.

I started the test suite.

ekohl:
Could you address the test failures?

spec/classes/dns_init_spec.rb
```diff
@@ -65,6 +65,15 @@
         end
       end
 
+      let(:defaultzonepath) do
+        case facts[:os]['family']
+        when 'Debian'
+          "#{etc_directory}/named.conf.default-zones"
+        when 'RedHat'
+          'unmanaged'
```
ekohl, line 73: I'd leave this out and only deal with absolute paths (or nil) in tests.
```suggestion
```

spec/classes/dns_init_spec.rb
```diff
@@ -207,34 +235,68 @@
...
+              '// named.conf',
+              "include \"#{rndc_key}\";",
+              'controls  {',
+              '        inet 127.0.0.1 port 953 allow { 127.0.0.1; } keys { "rndc-key"; };',
+              '};',
+              'options  {',
+              "        include \"#{options_path}\";",
+              '};',
+              localzonepath ? "include \"#{localzonepath}\";" : nil,
+              defaultzonepath ? "include \"#{defaultzonepath}\";" : nil,
```
ekohl, line 279: Rather than duplicating the whole block you could also change this to
```suggestion
              (defaultzonepath && defaultzonepath != 'unmanaged') ? "include \"#{defaultzonepath}\";" : nil,
```

Or you can drop the `unmanaged` bit.

benjamin-robertson: Switch to nil. No longer required

spec/classes/dns_init_spec.rb
benjamin-robertson: Done, updated.

benjamin-robertson:
@ekohl Tests should be passing now if you want to kick them off again in this repo. Thanks Ben

spec/classes/dns_init_spec.rb
```diff
@@ -140,9 +147,11 @@
             "        include \"#{options_path}\";",
             '};',
             "include \"#{localzonepath}\";",
+            defaultzonepath ? "include \"#{defaultzonepath}\";" : nil,
```
ekohl, line 150: I like this solution more than what was used for `localzonepath`.

ekohl (approved):
Thanks!

PR #218, opened 2022-09-13
Remove dnssec-enable option for EL >= 9

ikonia:
validated against EL (Redhat / CentOS / Rocky) 7 - 8 - 9 functions correctly.
Basic test against Fedora 35, which also works

Adds check of EL version greater than 7 and sets dnssec_enable parameter to undef if greater to allow the removal of the parameter from options.conf, which is a deprecated parameter in EL8 and later. This removes any warning in 9.15 and any breakage in 9.18 or later. Any version EL 7 or earlier includes the parameter allowing the user to set it on and off depending on if they want to use it, but the parameter is included as a valid parameter

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -54,8 +54,7 @@ class dns::params {
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = $facts['os']['Family'] ? {
-        'RedHat' => if versioncmp($facts['os']['release']['major'], '8') >= 0 { undef } else { 'yes' },
+      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' },
         default  => undef,
       }
     }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -54,7 +54,7 @@ class dns::params {
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' },
+      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' }
         default  => undef,
       }
     }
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -54,7 +54,10 @@
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = 'yes'
+      $dnssec_enable = $facts['os']['Family'] ? {
+        'RedHat' => if versioncmp($facts['os']['release']['major'], '8') >= 0 { undef } else { 'yes' },
+        default  => undef,
+      }
```
ekohl, line 60: You're already in the `RedHat` os family branch.
```suggestion
      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '8') >= 0 { undef } else { 'yes' }
```

However, I do think that on EL8 it's still relevant. That has 9.11 and only 9.15 made it obsolete. So should it be this?
```suggestion
      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' }
```

ikonia:
excellent feedback, happy to make the changes, I got the warning on my CentOS 8 bind host though, so I will go back and double check the version on that.

manifests/params.pp
ikonia: I'd not fully realised that higher up I'd joined the RedHat branch, this was a great spot and helpful, thank you

manifests/params.pp
```diff
@@ -54,7 +54,8 @@
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = 'yes'
+      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' }
+      }
```
ekohl, line 58:
```suggestion
```

ikonia:
do I need to look at the rspect test suites on this as while the module works fine on EL9 (which is where the test suite is failing) I'm not sure if these tests are known to fail, or if I should dig into them and look to update them. 
If they are expected passes I can look into updating them

ekohl:
Yes, the failures are caused by your change.

ikonia:
tests updated, thanks for the help

ekohl (approved):
Thanks!

PR #221, opened 2022-09-27
Adding scl based installation mode

diLLec:
The current module does not provide a way to deviate from the OS bind version. Therefore one is restricted to the version the OS is providing.

With the isc-bind SCL packages however, a seperated installation of bind can be used, based on the newest version of the software.

However this can only be done by modifying the default paths of configuration and zone data. 

This pull requests adds code to support those changed paths and adds a SCL overlay for the dig and nsupdate tools.


README.md
```diff
@@ -73,6 +73,22 @@ dns::logging_channels:
     syslog_facility: 'auth'
 ```
 
+## isc-bind Software Collections usage
+If you plan to use the isc-bind SCL packages (https://copr.fedorainfracloud.org/coprs/isc/bind/) the switch 'dns::globals::scl'
+must be set.
+
+This can be accomplished by using hiera
+```yaml
+  dns::params::scl: true
```
ekohl, line 82: No need to indent and I think you also need to set it on globals
```suggestion
dns::globals::scl: true
```

README.md
```diff
@@ -73,6 +73,22 @@ dns::logging_channels:
...
+``
+
+or class instantiation.
+```puppet
+  class { 'dns::params':
+    scl => true
+  }
```
ekohl, line 89:
```suggestion
```puppet
class { 'dns::params':
  scl => true,
}
```

manifests/globals.pp
```diff
@@ -0,0 +1,7 @@
+
+# @param $scl Enable the usage of SCL (see readme) when considering paths etc.
+#
+
```
ekohl, line 4: Whitespace around the comments breaks puppet-strings. Please also include a summary tag.
```suggestion
# @summary SUMMARY HERE
#
# @param $scl Enable the usage of SCL (see readme) when considering paths etc.
```

manifests/init.pp
```diff
@@ -181,6 +183,7 @@
   Hash[String, Hash] $keys                                          = {},
   Hash[String, Hash] $logging_categories                            = {},
   Hash[String, Hash] $logging_channels                              = {},
+  Boolean $redhat_scl                                               = false
```
ekohl, line 186: This looks unused now.

manifests/params.pp
```diff
@@ -32,29 +32,61 @@
...
+        $dns_server_package = 'bind'
+        $namedservicename = 'named'
+        $user = 'named'
+        $group = 'named'
+        $rndcconfgen = '/usr/sbin/rndc-confgen'
+        $named_checkconf = '/usr/sbin/named-checkconf'
+        $sysconfig_file = '/etc/sysconfig/named'
+        $sysconfig_template = "dns/sysconfig.${facts['os']['family']}.erb"
+        $sysconfig_startup_options = undef
+        $sysconfig_disable_zone_checking = undef
```
ekohl, line 83: Please keep the ones that are the same both both branches outside of the if/else block.

templates/scl_util.epp
```diff
@@ -0,0 +1,5 @@
+<%- | String $util,
+      String $sclenvname
+| -%>
+#!/bin/bash
+scl enable <%= $sclenvname %> -- <%= $util %> $@
```
ekohl, line 5: In other scripts I've used `scl_source` and then `exec`. The benefit is that you have fewer processes. While most utils are short lived, it really makes you `ps auxf` easier to read. 
```suggestion
. scl_source enable <%= $sclenvname %>
exec <%= $util %> "$@"
```

ekohl (changes requested):
The Foreman project is mostly moving away from EL7, but I'd be OK with adding this since it doesn't really interfere with other operations. It would be nice to add some verification in the form of an acceptance test.

manifests/install.pp
```diff
@@ -13,4 +13,24 @@
       require => $pkg_req,
     }
   }
+
+  if $facts['os']['family'] in ['RedHat'] and $dns::globals::scl {
+    ['dig', 'nsupdate', 'rdnc'].each | $util | {
+      file { "/usr/bin/${util}":
```
ekohl, line 19: Should this also require `$pkg_req`? If the package somehow failed to install, the wrappers are rather useless.

manifests/params.pp
```diff
@@ -32,29 +32,55 @@
...
       $sysconfig_disable_zone_checking = undef
 
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' }
+      if $dns::globals::scl {
+        $sclroot = '/opt/isc/isc-bind/root'
+        $sclconfroot = '/etc/opt/isc/scls/isc-bind'
+        $sclenvname = 'isc-bind'
```
ekohl, line 47: Perhaps I'm making it too dynamic, but consider:
```suggestion
        $sclvendor = 'isc'
        $sclenvname = "${sclvendor}-bind"
        $sclroot = "/opt/${sclvendor}/${sclenvname}/root"
        $sclconfroot = "/etc/opt/${sclvendor}/scls/${sclenvname}"
```

You could then even consider changing `$dns::globals::scl` to `Optional[Pattern[/^.+-.+$/]] $scl = undef`. You then only need to extract `$sclvendor` from `$scl` by splitting the string and it becomes possible to use different SCL too. Not sure if they exist so feel free to ignore that.

manifests/install.pp
diLLec: You are right - changed

manifests/params.pp
```diff
@@ -32,29 +32,55 @@
...
+
+        $dnsdir = $sclconfroot
+        $vardir = '/var/named'
```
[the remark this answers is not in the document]
diLLec: Absolutely right  - changed.

manifests/params.pp
```diff
@@ -32,29 +32,55 @@
...
+        $optionspath = "${dnsdir}/options.conf"
+        $zonefilepath = "${vardir}/dynamic"
+        $localzonepath = "${dnsdir}/named.rfc1912.zones"
+        $defaultzonepath = 'unmanaged'
+        $publicviewpath = "${dnsdir}/zones.conf"
+        $viewconfigpath = "${dnsdir}/views"
+        $dns_server_package = 'isc-bind'
```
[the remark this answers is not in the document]
diLLec: actually I would like to stick to the metapackage as it will install more dependencies which isc-bind-bind will not: 

```
$ yum install isc-bind
...
Installing:
 isc-bind
Installing dependencies:
 checkpolicy
 environment-modules
 isc-bind-bind
 isc-bind-bind-libs
 isc-bind-bind-utils
 isc-bind-fstrm
 isc-bind-fstrm-utils
 isc-bind-libevent
 isc-bind-libuv
 isc-bind-protobuf-c
 isc-bind-runtime
 jemalloc
 policycoreutils-python-utils
 python3-audit
 python3-libsemanage
 python3-policycoreutils
 python3-setools
 scl-utils
 tcl
 
 
--------------------------------------
$ yum install isc-bind-bind
...
Installing:             
 isc-bind-bind          
Installing dependencies:
 environment-modules    
 isc-bind-bind-libs     
 isc-bind-fstrm         
 isc-bind-fstrm-utils   
 isc-bind-libevent      
 isc-bind-libuv         
 isc-bind-protobuf-c    
 isc-bind-runtime       
 jemalloc               
 scl-utils              
 tcl                    
 ```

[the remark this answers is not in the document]
Comment (diLLec) [reply manifests/params.pp:57]: is that OK with u?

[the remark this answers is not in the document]
Comment (ekohl) [reply manifests/params.pp:57]: Perhaps then make it explicit what you're doing (and why). Something like
```suggestion
        # Install the SCL meta package to pull in utils
        $dns_server_package = $sclenvname
```

manifests/install.pp
```diff
@@ -13,4 +13,18 @@
...
+        owner   => root,
+        group   => root,
+        mode    => '0755',
+        content => epp('dns/scl_util.epp', { 'util' => $util, 'sclenvname' => $dns::sclenvname }),
+        require => $pkg_req
+      }
+    }
+  }
 }
+
```
ekohl, line 30:
```suggestion
```

manifests/params.pp
```diff
@@ -32,29 +32,58 @@
...
       $sysconfig_startup_options = undef
       $sysconfig_disable_zone_checking = undef
 
       # This option is not relevant for RedHat
       $sysconfig_resolvconf_integration = undef
 
-      $dnssec_enable = if versioncmp($facts['os']['release']['major'], '9') >= 0 { undef } else { 'yes' }
+      if $dns::globals::scl {
+        $sclprovider = $dns::globals::scl.split('-')[0]
+        $sclenvname = $dns::globals::scl
```
ekohl, line 47: Perhaps skip this variable now and only use `$scl`? Since `params` inherits from `globals` you can just use `$scl`, though I've seen that people have difficulty with that so I wouldn't mind being explicit in it.

manifests/config.pp
```diff
@@ -87,4 +87,15 @@
...
   }
+
+  # the RedHat SCL packages do not consist of the default rfc1912 zones file
+  # therefore the file is created out of this module
+  if $facts['os']['family'] in ['RedHat'] and $dns::globals::scl {
+    file { "${dns::params::localzonepath}":
+      owner   => root,
+      group   => root,
+      mode    => '0755',
+      content => template('dns/named.rfc1912.zones.erb'),
```
ekohl, line 98: Since it's an empty template, you could reduce it to:
```suggestion
      content => '',
```

Or in `params.pp` set it to `'unmanaged'` to skip the file altogether. That would be cleaner IMHO.

diLLec: actually this template should not be empty - I've added the content in the last commit.

manifests/params.pp
diLLec: split variant has been implemented

manifests/params.pp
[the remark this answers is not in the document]
diLLec: allright - change has been added

manifests/config.pp
```diff
@@ -87,4 +87,15 @@
       content => template($dns::sysconfig_template),
     }
   }
+
+  # the RedHat SCL packages do not consist of the default rfc1912 zones file
+  # therefore the file is created out of this module
+  if $facts['os']['family'] in ['RedHat'] and $dns::globals::scl {
+    file { "${dns::params::localzonepath}":
```
ekohl, line 94: No need for string interpolation. It's also better to refer to the main class rather than the private `params` class. This means it actually respects the parameter. Then you should make sure you check it's not `unmanaged`.
```suggestion
    file { $dns::localzonepath:
```

manifests/config.pp
```diff
@@ -87,4 +87,15 @@
       content => template($dns::sysconfig_template),
     }
   }
+
+  # the RedHat SCL packages do not consist of the default rfc1912 zones file
+  # therefore the file is created out of this module
+  if $facts['os']['family'] in ['RedHat'] and $dns::globals::scl {
```
ekohl, line 93:
```suggestion
  if $facts['os']['family'] == 'RedHat' and $dns::globals::scl {
```

manifests/globals.pp
```diff
@@ -0,0 +1,7 @@
+# @summary global default settings for the module that also can
+#
+# @param $scl Enable the usage of SCL (see readme) when considering paths etc.
+#
+class dns::globals (
+  Optional[Pattern[/^[a-z]+-[a-z]+$/]] $scl       = undef
```
ekohl, line 6:
```suggestion
  Optional[Pattern[/^[a-z]+-[a-z]+$/]] $scl = undef
```

manifests/config.pp
ekohl: Given it's a static file now, please move it to `files` and use:
```suggestion
      content => file('dns/named.rfc1912.zones'),
```

ekohl (commented):
I think this looks pretty good. Some small comments inline.

diLLec:
OK @ekohl thanks for your review - I've worked through your comments and applied the configuration to my test system. It all works with a happy isc-named with a very current version :). 

Does the acceptance test run after you acked my changes?

manifests/config.pp
```diff
@@ -87,4 +87,23 @@
...
+      owner   => root,
+      group   => root,
+      mode    => '0755',
+      content => file('dns/named.rfc1912.zones'),
+    }
+    ['named.loopback', 'named.empty', 'named.localhost'].each |$localzone_file| {
+      file { "${dns::vardir}/$localzone_file":
+        owner   => $dns::user,
+        group   => $dns::group,
+        mode    => '0755',
```
ekohl, line 104: Why are these files executable?
```suggestion
        mode    => '0644',
```

spec/acceptance/scl_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+if fact('osfamily') == 'RedHat'
+  describe 'Scenario: install scl-bind' do
+
+    it_behaves_like 'an idempotent resource' do
+      let(:manifest) do
+        <<-EOS
+        yumrepo{'bind_scl':
+          descr    => 'copr repo for bind owned by isc'
+          baseurl  =>  "https://download.copr.fedorainfracloud.org/results/isc/bind/epel-$facts['os']['release']['major']-x86_64/"
+          gpgcheck =>  0
```
ekohl, line 12: I think you're missing commas here
```suggestion
          descr    => 'copr repo for bind owned by isc',
          baseurl  => "https://download.copr.fedorainfracloud.org/results/isc/bind/epel-$facts['os']['release']['major']-x86_64/",
          gpgcheck => 0,
```

spec/acceptance/scl_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper_acceptance'
+
+if fact('osfamily') == 'RedHat'
```
ekohl, line 3: Does this work on EL8 and EL9 or should it be limited to EL7?

spec/acceptance/scl_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper_acceptance'
```
ekohl, line 1: My biggest worry about this test is that it's silently running with the non-SCL version. There is no cleanup before or after. Perhaps it needs a `before(:context)` and `after(:context)` where it removes both the SCL and non-SCL versions.

diLLec: good point - I've added some code for this but I'm very unvertain if it is correct. Could you check please?

spec/acceptance/dns_spec.rb
```diff
@@ -29,4 +29,17 @@
   describe command('dig +short SOA example.com @localhost') do
     its(:stdout) { is_expected.to match("ns1.example.com. root.example.com. 1 86400 3600 604800 3600\n") }
   end
+
+  before(:context) do
+    describe command('dnf remove -y bind bind-utils') do
+      its(:stdout) { should match /Removed:/ }
+    end
```
ekohl: This assumes dnf is present, but it's not everywhere. I think you can use:
```suggestion
    on hosts, puppet('resource', 'package', 'bind', 'ensure=purged')
    on hosts, puppet('resource', 'package', 'bind-utils', 'ensure=purged')
```

spec/acceptance/dns_spec.rb
```diff
@@ -29,4 +29,17 @@
...
+  end
+
+  after(:context) do
```
ekohl: Shouldn't this have been in the SCL acceptance spec file?

ekohl (commented):
Sorry, this got a bit lost.

ekohl:
I apologize for never getting around to actually merging this, but given the EL7 EOL I'm going to drop the support in the next major version via https://github.com/theforeman/puppet-dns/pull/262.

PR #222, opened 2022-10-05
changed support IRC server from Freenode to Libera

ikonia:
very minor change to readme to list correct IRC server network

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ See the CONTRIBUTING.md file for much more information.
 
 # More info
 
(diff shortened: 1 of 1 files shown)


evgeni:
While the change is correct, I wonder if we should mention IRC here at all, vs just saying "go to foreman community".

@ehelms @ekohl thoughts?

ekohl:
Should we link to https://theforeman.org/support.html instead?

ikonia:
is that the consensus ? change the link to the support link rather than the correct IRC link ?

evgeni:
Yes, please

ikonia:
done

ekohl (approved):
Thanks!

PR #224, opened 2022-11-07
logging channel error message typo faility to facility

ikonia:
minor typo in syslog_facility error message from channel.pp detailing incorrect parameter


ekohl (approved):
Thanks!

PR #225, opened 2022-12-26
Remove legacy fact usage

smortex:
We can now run puppet without legacy facts:

```
puppet config set include_legacy_facts false
```

When doing so, the module fail to get the default value for `$dns::zone::soa`.  Fix it by using the `networking.fqdn` fact.

While here, adjust a few more legacy facts.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -63,7 +63,7 @@
 define dns::zone (
   Array[String] $target_views                             = [],
   String $zonetype                                        = 'master',
-  String $soa                                             = $fqdn,
+  String $soa                                             = $facts['networking']['fqdn'],
   Boolean $reverse                                        = false,
   String $ttl                                             = '10800',
   Optional[Stdlib::IP::Address::V4] $soaip                = undef,
--- a/spec/defines/dns_logging_category_spec.rb
+++ b/spec/defines/dns_logging_category_spec.rb
@@ -3,7 +3,7 @@ require 'spec_helper'
 describe 'dns::logging::category' do
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
-      let(:facts) { os_facts.merge(fqdn: 'puppetmaster.example.com') }
+      let(:facts) { override_facts(os_facts, networking: { fqdn: 'puppetmaster.example.com' }) }
       let(:pre_condition) { 'include dns' }
       let(:title) { 'test' }
       let(:logdir) { '/var/log/named' }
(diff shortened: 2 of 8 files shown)


manifests/zone.pp
```diff
@@ -63,7 +63,7 @@
 define dns::zone (
   Array[String] $target_views                             = [],
   String $zonetype                                        = 'master',
-  String $soa                                             = $fqdn,
+  String $soa                                             = $facts['networking']['fqdn'],
```
ekohl, line 66: Odd that the lint plugin to find legacy facts didn't find this.

templates/options.conf.erb
```diff
@@ -26,7 +26,7 @@ listen-on-v6 { <%= scope.lookupvar('::dns::listen_on_v6') %>; };
 allow-recursion { <%= scope.lookupvar('::dns::allow_recursion').join("; ") %>; };
 <% end -%>
 
-<% if (@osfamily =~ /^(FreeBSD|DragonFly)$/) -%>
+<% if (@os['family'] =~ /^(FreeBSD|DragonFly)$/) -%>
```
ekohl, line 29: I'd prefer to use the facts hash rather than relying on top level variables.
```suggestion
<% if (@facts['os']['family'] =~ /^(FreeBSD|DragonFly)$/) -%>
```

spec/defines/dns_logging_category_spec.rb
```diff
@@ -3,7 +3,7 @@
 describe 'dns::logging::category' do
   on_supported_os.each do |os, os_facts|
     context "on #{os}" do
-      let(:facts) { os_facts.merge(fqdn: 'puppetmaster.example.com') }
+      let(:facts) { os_facts.merge(networking: { fqdn: 'puppetmaster.example.com' }) }
```
ekohl, line 6: Please use [`override_facts` from voxpupuli-test](https://github.com/voxpupuli/voxpupuli-test#fact-handling) for this (otherwise there are no other networking facts):
```suggestion
      let(:facts) { override_facts(os_facts, networking: { fqdn: 'puppetmaster.example.com' }) }
```

ekohl (changes requested):
Odd that

spec/defines/dns_logging_category_spec.rb
smortex: Nice!

ekohl (approved):

ekohl:
I'll consider this as a bugfix to 6426a394af5afebe7fe0c0c0f2b834198c75a52b.

PR #231, opened 2023-05-03
Drop Debian 9 & Fedora 32 and add Fedora 37/38

ekohl:


evgeni:
What's the point of Fedora support here (= in this module)?

ekohl:
I think it's still useful for future versions, though we should do a better job of keeping it up to date.

PR #232, opened 2023-05-03
Refs #36345 - Drop Puppet 6 support

ekohl:


ekohl:
https://github.com/theforeman/puppet-dns/pull/231 drops Debian 9 support. Together they should be green.

Issue #234, opened 2023-06-28
sysconfig_file for Debian and Ubuntu is wrong and has long changed to /etc/default/named


frittentheke:
Currently this module still writes to `/etc/default/bind9`, but since Debian Bullseye and at least Ubuntu Focal (20.04) the defaults was renamed to `/etc/default/named`.

See https://salsa.debian.org/dns-team/bind9/-/blob/debian/9.18/debian/bind9.named.service#L10
unfortunately there was never a note in the release notes (https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=%23993257), but still the current value is wrong.``

ekohl:
Thanks for pointing this out. Could you come up with a PR to resolve this?

PR #235, opened 2023-06-28
Change sysconfig_file for Debian/Ubuntu to '/etc/default/named'

frittentheke:
This was changed quite a while back, see https://salsa.debian.org/dns-team/bind9/-/commit/6fd962a36bb54e0106e001aff6cce056d54e2526.

Fixes https://github.com/theforeman/puppet-dns/issues/234

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -20,7 +20,11 @@ class dns::params {
         'Ubuntu' => if versioncmp($facts['os']['release']['major'], '22.04') >= 0 { '/usr/bin/named-checkconf' } else { '/usr/sbin/named-checkconf' },
         default  => '/usr/sbin/named-checkconf',
       }
-      $sysconfig_file     = '/etc/default/bind9'
+      $sysconfig_file     = $facts['os']['name'] ? {
+        'Debian' => '/etc/default/bind9',
+        'Ubuntu' => if versioncmp($facts['os']['release']['major'], '20.04') >= 0 { '/etc/default/named' } else { '/etc/default/bind9' },
+        default  => '/etc/default/named',
+      }
       $sysconfig_template = "dns/sysconfig.${facts['os']['family']}.erb"
       $sysconfig_startup_options = '-u bind'
       $sysconfig_resolvconf_integration = false
--- a/spec/classes/dns_init_spec.rb
+++ b/spec/classes/dns_init_spec.rb
@@ -403,13 +403,15 @@ describe 'dns' do
         }
       end
 
-      context 'sysconfig', if: ['Debian', 'RedHat'].include?(os_facts[:os]['family']) do
+      context 'sysconfig', if: ['Debian', 'RedHat', 'Ubuntu'].include?(os_facts[:os]['family']) do
         let(:sysconfig_named_path) do
           case facts[:os]['family']
           when 'RedHat'
             '/etc/sysconfig/named'
           when 'Debian'
             '/etc/default/bind9'
+          when 'Ubuntu'
+            facts[:os]['release']['major'] == '18.04' ? '/etc/default/bind9' : '/etc/default/named'
           end
         end
 


manifests/params.pp
```diff
@@ -20,7 +20,7 @@
         'Ubuntu' => if versioncmp($facts['os']['release']['major'], '22.04') >= 0 { '/usr/bin/named-checkconf' } else { '/usr/sbin/named-checkconf' },
         default  => '/usr/sbin/named-checkconf',
       }
-      $sysconfig_file     = '/etc/default/bind9'
+      $sysconfig_file     = '/etc/default/named'
```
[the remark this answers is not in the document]
frittentheke: I now added Debian 10 and Ubuntu 18.04 as exceptions ... I hope I did this right.

frittentheke:
@ekohl could you PTAnotherL :-)

manifests/params.pp
```diff
@@ -20,7 +20,11 @@
         'Ubuntu' => if versioncmp($facts['os']['release']['major'], '22.04') >= 0 { '/usr/bin/named-checkconf' } else { '/usr/sbin/named-checkconf' },
         default  => '/usr/sbin/named-checkconf',
       }
-      $sysconfig_file     = '/etc/default/bind9'
+      $sysconfig_file     = $facts['os']['name'] ? {
+        'Debian' => if versioncmp($facts['os']['release']['major'], '10') >= 0 { '/etc/default/bind9' } else { '/etc/default/named' },
+        'Ubuntu' => if versioncmp($facts['os']['release']['major'], '18.04') >= 0 { '/etc/default/bind9' } else { '/etc/default/named' },
```
ekohl, line 25: Same here. I think your logic is incorrect and should be:
```suggestion
        'Ubuntu' => if versioncmp($facts['os']['release']['major'], '20.04') >= 0 { '/etc/default/named' } else { '/etc/default/bind9' },
```

manifests/params.pp
```diff
@@ -20,7 +20,11 @@
         'Ubuntu' => if versioncmp($facts['os']['release']['major'], '22.04') >= 0 { '/usr/bin/named-checkconf' } else { '/usr/sbin/named-checkconf' },
         default  => '/usr/sbin/named-checkconf',
       }
-      $sysconfig_file     = '/etc/default/bind9'
+      $sysconfig_file     = $facts['os']['name'] ? {
+        'Debian' => if versioncmp($facts['os']['release']['major'], '10') >= 0 { '/etc/default/bind9' } else { '/etc/default/named' },
```
ekohl, line 24: We only support Debian  >= 10 so should this be simple:
```suggestion
        'Debian' => '/etc/default/bind9',
```

But it's actually not what reflects reality. I think you meant this:
```suggestion
        'Debian' => if versioncmp($facts['os']['release']['major'], '11') >= 0 { '/etc/default/named' } else { '/etc/default/bind9' },
```

ekohl (changes requested):
I was surprised the test for the default file doesn't require an update, but I wonder if your logic is correct. It suggests it's actually a noop compared to the current code for all supported versions.

manifests/params.pp
frittentheke: pushed a change to fix this.

manifests/params.pp
frittentheke: thanks for the simplification, I pushed a change.

frittentheke:
@ekohl could you please take another look?

spec/classes/dns_init_spec.rb
```diff
@@ -403,13 +403,15 @@
...
-      context 'sysconfig', if: ['Debian', 'RedHat'].include?(os_facts[:os]['family']) do
+      context 'sysconfig', if: ['Debian', 'RedHat', 'Ubuntu'].include?(os_facts[:os]['family']) do
         let(:sysconfig_named_path) do
           case facts[:os]['family']
           when 'RedHat'
             '/etc/sysconfig/named'
           when 'Debian'
             '/etc/default/bind9'
+          when 'Ubuntu'
+            '/etc/default/named'
```
ekohl, line 414: I think this should make the tests pass:
```suggestion
            facts[:os]['release']['major'] == '18.04' ? '/etc/default/bind9' : '/etc/default/named'
```

frittentheke: Thanks for the help. I am not really familiar with Ruby and the testing / mocking here.
Unfortunately this was not "it", yet ...

ekohl:
The problem in the tests was that you were confusing the OS family and name facts. You also missed that Debian 11+ has changed it. So I ended up submitting https://github.com/theforeman/puppet-dns/pull/238 for this.

PR #238, opened 2023-08-28
Change sysconfig_file for Debian/Ubuntu to '/etc/default/named'

### LINKED ISSUE #234: sysconfig_file for Debian and Ubuntu is wrong and has long changed to /etc/default/named

ekohl:
It is was changed quite a [while back](https://salsa.debian.org/dns-team/bind9/-/commit/6fd962a36bb54e0106e001aff6cce056d54e2526).

Fixes #234

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -20,7 +20,11 @@ class dns::params {
         'Ubuntu' => if versioncmp($facts['os']['release']['major'], '22.04') >= 0 { '/usr/bin/named-checkconf' } else { '/usr/sbin/named-checkconf' },
         default  => '/usr/sbin/named-checkconf',
       }
-      $sysconfig_file     = '/etc/default/bind9'
+      $sysconfig_file     = $facts['os']['name'] ? {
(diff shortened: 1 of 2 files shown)


frittentheke:
@evgeni did you get a chance to look at this one?

evgeni:
If only GitHub would notify people about a requested review… Looking.

evgeni (approved):
one could argue we could just drop support for those old releases, but until we do the diff is correct

ekohl:
> one could argue we could just drop support for those old releases, but until we do the diff is correct

I didn't want to do a breaking change just now.

PR #244, opened 2023-11-21
Add support for empty forwarders in master zone

mbarecki:
This gives possibility to disable forwarding for delegated zone when server is authoritative for domain and global forwarding is set.

### DIFF
--- a/manifests/zone.pp
+++ b/manifests/zone.pp
@@ -51,6 +51,9 @@
 # @param zonefilepath
 # @param filename
 # @param forward
+# @param master_empty_forwarders_enable
+#   Enable empty forwarders option in master zone.
+#   Applicable when forwarders are empty and zonetype is set to master.
 # @param forwarders
 # @param dns_notify
 # @param key_directory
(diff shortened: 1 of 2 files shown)


manifests/zone.pp
```diff
@@ -51,6 +51,8 @@
 # @param zonefilepath
 # @param filename
 # @param forward
+# @param master_empty_forwarders_enable
+#   Enable empty forwarders option in master zone
```
ekohl, line 55: I think this is a bit brief. It doesn't explain that `forwarders` must be empty and `zonetype` set to `master`.

templates/named.zone.erb
```diff
@@ -10,6 +10,9 @@ zone "<%= @zone %>" {
     file "<%= @zonefilename %>";
 <% end -%>
 <% if @zonetype == 'master' -%>
+<% if @forwarders.empty? && @master_empty_forwarders_enable -%>
+    forwarders {};
```
ekohl, line 14: Is there a reason to enforce that `forwarders` is empty? Isn't there a case to be made for `master_forwarders_enable`?

https://serverfault.com/questions/713256/purpose-of-a-forwarder-option-in-a-master-zone suggests there is also a case to be made to actually include forwarders, instead of discarding the `forwarders` values provided.

manifests/zone.pp
mbarecki: Hello @ekohl,

Thank you for looking into this and for your feedback.

Description for `master_empty_forwarders_enable` parameter was updated as you suggested.

templates/named.zone.erb
mbarecki: Possibility to enforce empty forwarders is for disabling global forwarding when server is authoritative for domain and subdomain is delegated.
Related link:
https://kb.isc.org/docs/aa-00538

Possibility to include forwarders in master zone was introduced now as well.

ekohl (approved):

PR #246, opened 2024-01-30
Add dns::dnssec_keygen function

ekohl:
This wraps dnssec-keygen and returns the data in a structured format. The result can be used to secure communications. It always generates a random key and the resulting data should be cached or otherwise made idempotent.


ekohl:
This no longer relates to the Redmine issue because that needs `tsig_keygen` (https://github.com/theforeman/puppet-dns/pull/253). I still think this can be useful to some.

evgeni (approved):

PR #249, opened 2024-03-26
Update ensure_packages->stdlib::ensure_packages; require stdlib 9

bastelfreak:
The `ensure_packages()` function is deprecated with stdlib 9. It triggers a catalog compilation warning on Puppet 7 and a catgalog compilation error on Pupppet 8. This PR switches to the successor, `stdlib::ensure_packages()`. This function got introduced in stdlib 9.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -2,7 +2,7 @@
 # @api private
 class dns::install {
   if ! empty($dns::dns_server_package) {
-    ensure_packages([$dns::dns_server_package])
+    stdlib::ensure_packages([$dns::dns_server_package])
     $pkg_req = Package[$dns::dns_server_package]
   } else {
     $pkg_req = undef
--- a/metadata.json
+++ b/metadata.json
@@ -22,7 +22,7 @@
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 10.0.0"
+      "version_requirement": ">= 9.0.0 < 10.0.0"
     }
   ],
   "requirements": [


evgeni:
This should then update the minimum version of stdlib to 9?
But also, I thought the non namespaced versions are supposed to still work?!

bastelfreak:
They do work, but you will see this on Puppet 7:
```
2024-03-26T13:00:16.305+01:00 WARN  [qtp1279219621-121] [puppetserver] Puppet This function is deprecated, please use stdlib::ensure_packages instead. at ["/etc/puppetlabs/code/environments/production/modules/dns/manifests/install.pp", 5]:["/etc/puppetlabs/code/environments/production/modules/dns/manifests/init.pp", 185]
```

And on Puppet 8 it's an error (and if you switch strict mode to on on puppet 7, it will fail there as well). I suggest to merge #250 for a new minor releae, then merge #249 and do a major release.

evgeni:
Why does CI then pass on Puppet 8 without that change?

bastelfreak:
Apparently this is a whole clusterfuck that broke so much stuff that there's now a patch within the deprecation method:
https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/deprecation.rb#L27-L35

And the old `ensure_packages()` hacks around the `fail()` call by passing `false` as last argument to `deprecation()`:

https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/lib/puppet/functions/ensure_packages.rb#L10

This got implemented in stdlib 9.something. But 9.0.0 + Puppet 8 will actually fail.

ekohl:
If only there had been a release of Puppet 8 where the methods were deprecated **and** replacements available. That would have made the whole migration so much less painful.

ekohl (approved):
In other places we've treated this as a major version bump, given the pain of stdlib 9. At this point I'm not too worried about it anymore given our other modules have made the jump too.

evgeni:
So we just forbid 9.x < 9.2.0 and are done?

bastelfreak:
> So we just forbid 9.x < 9.2.0 and are done?

We had that idea but if I remember it correctly metadata.json doesn't support that properly. Also the whole hack in stdlib is a workaround and ideally we move away from the old functions.

Issue #251, opened 2024-05-13
Add `response-policy` to the `dns::view`


supun983:
Could you please add response-policy to the `dns::view` defined type, so it can be called from view.pp or else, simillar option like `additional_directives` and `additional_options`.

**As an example**
```
view "example" {
  match-clients { 192.168.0.0/24; };
  response-policy {
    zone "rpz";
  };
  zone "rpz" {
    type master;
    file "/var/cache/bind/zones/db.rpz";
  };
};
```

Thank you.

ekohl:
A PR for this is welcome.

PR #252, opened 2024-05-16
Add response-policy to dns::view

supun983:
This PR, try to addresses the issue https://github.com/theforeman/puppet-dns/issues/251, which involves adding a response-policy to the `dns::view`.

Below example shows how to use the `dns::view` defined type with the `response-policy` parameter:
```
dns::view { 'example_view':
  # ...
  response_policy      => [
    {
      'zone'           => 'example.com',
      'policy'         => 'passthru',
      'max_policy_ttl' => 3600,
      'log'            => true,
    },
    {
      'zone'         => 'example.net',
      'policy'       => 'cname',
      'cname_domain' => 'example.net',
    },
  ],
}
```

It will generate the following `BIND` configuration:
```
view "example_view" {
    # ...
    response-policy {
        zone "example.com" policy passthru max-policy-ttl 3600 log true;
        zone "example.net" policy cname example.net;
    };
}
```

The `response-policy` may support a few other parameters, which I have not included for simplicity.

### DIFF
--- a/manifests/view.pp
+++ b/manifests/view.pp
@@ -38,6 +38,14 @@
 #    not to be unmanaged to be effective.
 # @param order
 #   The order parameter to the concat fragment.
+# @param response_policy
+#   Optional. An array of response policy configurations for the view in the
+#   following format:
+#   [{'zone' => '<ZONE_NAME>', 'policy' => '<POLICY_ACTION>', 'log' => true|false,
+#     'max_policy_ttl' => <TTL_VALUE>, 'cname_domain' => '<CNAME_DOMAIN>'}]
+#   Example: [{'zone' => 'example.com', 'policy' => 'passthru', 'log' => true,
+#              'max_policy_ttl' => 3600}, {'zone' => 'example.net',
+#              'policy' => 'cname', 'cname_domain' => 'example.com'}]
 #
 define dns::view (
   Array[String]                         $match_clients        = [],
@@ -57,6 +65,7 @@ define dns::view (
   Boolean                               $include_localzones   = true,
   Boolean                               $include_defaultzones = true,
   String                                $order                = '-',
+  Optional[Dns::ResponsePolicy]         $response_policy      = undef,
 ) {
   unless $dns::enable_views {
     fail('Must set $dns::enable_views to true in order to use dns::view')
--- a/types/responsepolicy.pp
+++ b/types/responsepolicy.pp
@@ -1,0 +1,12 @@
+type Dns::ResponsePolicy = Array[
+  Struct[{
+    zone           => String[1],
+    policy         => Optional[Enum[
+      'given', 'disabled', 'passthru', 'drop',
+      'nxdomain', 'nodata', 'tcp-only', 'cname'
+    ]],
+    cname_domain   => Optional[String[1]],
+    max_policy_ttl => Optional[Integer],
+    log            => Optional[Boolean]
+  }], 1, 32
+]
--- a/templates/named.view_header.erb
+++ b/templates/named.view_header.erb
@@ -41,6 +41,13 @@ view "<%= @title %>" {
 <% if @dnssec_validation -%>
     dnssec-validation <%= @dnssec_validation %>;
 <% end -%>
+<% if @response_policy -%>
+    response-policy {
+<% @response_policy.each do |policy| -%>
+        zone "<%= policy['zone'] %>"<% if policy['policy'] -%> policy <%= policy['policy'] %><% end -%><% if policy['policy'] == 'cname' && policy['cname_domain'] -%> <%= policy['cname_domain'] %><% end -%><% if policy['max_policy_ttl'] -%> max-policy-ttl <%= policy['max_policy_ttl'] %><% end -%><% if policy['log'] -%> log <%= policy['log'] %><% end -%>;
+<% end -%>
+    };
+<% end -%>
 
 <% if @include_localzones -%>
 <% if scope.lookupvar("::dns::localzonepath") != 'unmanaged' -%>


types/responsepolicy.pp
```diff
@@ -0,0 +1,12 @@
+type Dns::ResponsePolicy = Array[
+  Struct[{
+    zone           => String[1],
```
ekohl, line 3: I imagine `zone` is a domain, so we can enforce a stricter type
```suggestion
    zone           => Stdlib::Fqdn,
```

types/responsepolicy.pp
```diff
@@ -0,0 +1,12 @@
...
+    policy         => Optional[Enum[
+      'given', 'disabled', 'passthru', 'drop',
+      'nxdomain', 'nodata', 'tcp-only', 'cname'
+    ]],
+    cname_domain   => Optional[String[1]],
```
ekohl, line 8: Is this also a domain?
```suggestion
    cname_domain   => Optional[Stdlib::Fqdn],
```

types/responsepolicy.pp
```diff
@@ -0,0 +1,12 @@
...
+    max_policy_ttl => Optional[Integer],
```
ekohl, line 9: Can we tighten this with a lower bound like `0`?
```suggestion
    max_policy_ttl => Optional[Integer[0]],
```

manifests/view.pp
```diff
@@ -38,6 +38,14 @@
 #    not to be unmanaged to be effective.
 # @param order
 #   The order parameter to the concat fragment.
+# @param response_policy
+#   Optional. An array of response policy configurations for the view in the
+#   following format:
+#   [{'zone' => '<ZONE_NAME>', 'policy' => '<POLICY_ACTION>', 'log' => true|false,
+#     'max_policy_ttl' => <TTL_VALUE>, 'cname_domain' => '<CNAME_DOMAIN>'}]
```
ekohl, line 45: puppet-strings should link to the type alias, so explaining the format is probably redundant.

manifests/view.pp
```diff
@@ -38,6 +38,14 @@
...
+#   Example: [{'zone' => 'example.com', 'policy' => 'passthru', 'log' => true,
+#              'max_policy_ttl' => 3600}, {'zone' => 'example.net',
+#              'policy' => 'cname', 'cname_domain' => 'example.com'}]
```
ekohl, line 48: Perhaps add a full `@example` instead?

ekohl (commented):
Thanks for the PR. It looks good, just some minor comments.
It would be great if you could add a test or 2 in https://github.com/theforeman/puppet-dns/blob/master/spec/defines/dns_view_spec.rb.

types/responsepolicy.pp
supun983: It seems `Stdlib::Fqdn` might be too restrictive for the zone field
example: rpz
Using `String[1]` will allow more flexibility in specifying zone to allow both FQDN and non-FQDN values

types/responsepolicy.pp
supun983: yes, It should be a domain.

types/responsepolicy.pp
supun983: okay, no problem

manifests/view.pp
supun983: How about this
```
# @param response_policy
#   Optional. An array of response policy configurations for the view in the
#   following format:
# @example
#   dns::view { 'example':
#     response_policy => [
#       {'zone' => 'example', 'policy' => 'passthru', 'log' => true, 'max_policy_ttl' => 3600},
#       {'zone' => 'example.net', 'policy' => 'cname', 'cname_domain' => 'example.com'}
#     ],
#   }
#
```

supun983:
For the test case, I have added most of the possible combinations below. However, I am not much familliar with this :)

```
context "with response policy configuration" do
  let(:title) { "response_policy_view" }
  let :pre_condition do
    'class {"::dns": enable_views => true}'
  end
  let(:params) { {
    response_policy: [
      {'zone' => 'example.com', 'policy' => 'passthru', 'log' => true, 'max_policy_ttl' => 3600},
      {'zone' => 'rpz', 'policy' => 'drop'},
      {'zone' => 'example.net', 'policy' => 'cname', 'cname_domain' => 'example.com'},
      {'zone' => 'rpz.local', 'policy' => 'nxdomain'},
      {'zone' => 'example.org', 'policy' => 'nodata', 'log' => false},
      {'zone' => 'private', 'policy' => 'passthru'},
    ],
  } }

  it { should compile.with_all_deps }

  it "should have response policy configuration" do
    verify_concat_fragment_exact_contents(catalogue, 'dns_view_header_response_policy_view.dns', [
      'view "response_policy_view" {',
      '    response-policy {',
      '        zone "example.com" policy passthru max-policy-ttl 3600 log true;',
      '        zone "rpz" policy drop;',
      '        zone "example.net" policy cname example.com;',
      '        zone "rpz.local" policy nxdomain;',
      '        zone "example.org" policy nodata log false;',
      '        zone "private" policy passthru;',
      '    };',
    ])
  end
end
```

manifests/view.pp
ekohl: I think that's good, but `@example` takes a description of the example. This allows puppet-strings to correctly list multiple examples and link to them.

types/responsepolicy.pp
ekohl: Technically `Stdlib::Fqdn` allows `rpz`. It may be poorly named, but I'd prefer that. Strong validation is important to me.

supun983: yes, you are correct.

manifests/view.pp
supun983: Yes, a brief description is important. Further, I would like to add multiple examples for separate use cases like below.

```
# @param response_policy
#   Optional. An array of response policy configurations for the view.
#
# @example Complex response policy with multiple zones and policies
#   dns::view { 'complex_example':
#     response_policy => [
#       {
#         'zone' => 'example',
#         'policy' => 'passthru',
#         'log' => true,
#         'max_policy_ttl' => 3600
#       },
#       {
#         'zone' => 'example.net',
#         'policy' => 'cname',
#         'cname_domain' => 'example.com'
#       }
#     ],
#   }
#
# @example Simple response policy with a single zone
#   dns::view { 'simple_example':
#     response_policy => [
#       {
#         'zone' => 'rpz.example.com'
#       }
#     ],
#   }
```

supun983: Hi, 
kindly review this and please proceed if everything is okay

supun983:
Hi, please let me know how to proceed with the changes.

supun983:
The commit suggestion actions failed due to outdated changes and CI pipeline issues. I’ve added a test case to address the feedback as well. Could you please review it when you get a chance?

Issue #256, opened 2024-06-27
Please update the named.ca file


bigon:
Hello,

The root DNS file currently in the git repository is from 2008 (Feb 04, 2008)

The current one on internic is from June 26, 2024

I guess it should be updated

https://www.internic.net/domain/named.root

ekohl:
As mentioned in https://github.com/theforeman/puppet-dns/pull/257#pullrequestreview-2144972868, it's no longer used and should be removed.

PR #257, opened 2024-06-27
Update named.ca to version 2024062601

bigon:
Fixes: #256


ekohl (changes requested):
I completely didn't notice this repository had the CA file, so I started to look for where it was used. Then I noticed I removed its use in 7e9f8e1193b973fd8768fffeea13a166d06d1cd8, which is already 12 years ago. We should drop it instead.

bigon:
Oh right, TBH I didn't verified this

PR #258, opened 2024-06-27
Drop files/named.ca

bigon:
This file is not needed since 7e9f8e1193b973fd8768fffeea13a166d06d1cd8

Fixes: #256


ekohl (approved):
Thanks!

Issue #255, opened 2024-06-26
Add option to set disable-empty-zone


bigon:
Hello,

Would be nice to have an option to set `disable-empty-zone`

in that option needs to be set multiple time, for example:

```
disable-empty-zone "16.172.IN-ADDR.ARPA";
disable-empty-zone "17.172.IN-ADDR.ARPA";
disable-empty-zone "18.172.IN-ADDR.ARPA";
disable-empty-zone "19.172.IN-ADDR.ARPA";
disable-empty-zone "20.172.IN-ADDR.ARPA";
disable-empty-zone "21.172.IN-ADDR.ARPA";
disable-empty-zone "22.172.IN-ADDR.ARPA";
disable-empty-zone "23.172.IN-ADDR.ARPA";
disable-empty-zone "24.172.IN-ADDR.ARPA";
disable-empty-zone "25.172.IN-ADDR.ARPA";
disable-empty-zone "26.172.IN-ADDR.ARPA";
disable-empty-zone "27.172.IN-ADDR.ARPA";
disable-empty-zone "28.172.IN-ADDR.ARPA";
disable-empty-zone "29.172.IN-ADDR.ARPA";
disable-empty-zone "30.172.IN-ADDR.ARPA";
disable-empty-zone "31.172.IN-ADDR.ARPA";
```

PR #259, opened 2024-06-27
Add parameter to set disable-empty-zone option

bigon:
disable-empty-zone allows one to disable the automatically created zones

Fixes: #255

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -134,6 +134,9 @@
 #   A hash of logging categories to be created. See dns::logging::category for options.
 # @param logging_channels
 #   A hash of logging channels to be created. See dns::logging::channel for options.
+# @param disable_empty_zones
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -134,6 +134,9 @@
 #   A hash of logging categories to be created. See dns::logging::category for options.
 # @param logging_channels
 #   A hash of logging channels to be created. See dns::logging::channel for options.
+# @param disable_empty_zones
+# A hash containing a list of empty zones that shouldn't be created by bind
+# See: https://kb.isc.org/docs/aa-00800
```
ekohl, line 139:
```suggestion
#   A hash containing a list of empty zones that shouldn't be created by bind
#   See: https://kb.isc.org/docs/aa-00800
```

ekohl (changes requested):

ekohl (approved):
Thanks!

PR #260, opened 2024-06-27
Fixes #37604 - Validate DNS forwarders

ekohl:
A user can input an invalid value and the service will refuse to start up. We can catch this in data types, preventing service downtime.

### DIFF
--- a/types/forwarder.pp
+++ b/types/forwarder.pp
@@ -1,0 +1,11 @@
+# @summary a DNS forwarder entry
+#
+# A forwarder is an IP address (v4 or v6) with optionally followed a port.
+# Since we can't compose patterns, this copies stdlib's implementation for v4.
+# For v6 it uses the default type and grossly simplifies the port check for simplicity.
+type Dns::Forwarder = Variant[
+  Pattern[/\A([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(\s+port\s+[0-9]{1,5})?\z/],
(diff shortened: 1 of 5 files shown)


ehelms (approved):

types/forwarder.pp
```diff
@@ -0,0 +1,11 @@
...
+#
+# A forwarder is an IP address (v4 or v6) with optionally followed a port.
+# Since we can't compose patterns, this copies stdlib's implementation for v4.
+# For v6 it uses the default type and grossly simplifies the port check for simplicity.
+type Dns::Forwarder = Variant[
+  Pattern[/\A([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\.([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])){3}(\s+port\s+[0-9]{1,5})?\z/],
+  Stdlib::IP::Address::V6::Nosubnet,
+  # This is a really gross simplification of IPv6
+  Pattern[/(\A(:{0,2}[[:xdigit:]]{1,4}){1,8}\s+port\s[0-9]{1,5}\Z)/],
+]
```
evgeni, line 11: Technically, it can also be `<ip> port <int> tls <name>` and similar:
https://bind9.readthedocs.io/en/latest/reference.html#namedconf-statement-forwarders

Well, at least since 9.19.10: https://bind9.readthedocs.io/en/v9.19.10/notes.html#new-features

I don't think we need to care right now, but in the future

evgeni (approved):

types/forwarder.pp
ekohl: I didn't see that in `man named.conf` on Fedora 39

evgeni: Fedora still has 9.18.something

PR #262, opened 2024-07-15
Drop RHEL 7, CentOS 7 & 8, Scientific 7 & Debian 10; Add RHEL 9 & Fedora 39 & 40

ekohl:
These are end of life and testing on them isn't possible anymore.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -35,7 +35,8 @@
     {
       "operatingsystem": "RedHat",
       "operatingsystemrelease": [
-        "8"
+        "8",
+        "9"
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


metadata.json
```diff
@@ -35,43 +35,33 @@
...
+        "9"
       ]
     },
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
-        "37",
-        "38"
+        "39",
+        "40"
```
evgeni, line 52: I am still "why do we support fedora" camp, but :woman_shrugging:

evgeni (approved):

PR #265, opened 2024-08-08
Fix the documentation of the disable_empty_zones parameter

bigon:


manifests/init.pp
```diff
@@ -135,7 +135,7 @@
 # @param logging_channels
 #   A hash of logging channels to be created. See dns::logging::channel for options.
 # @param disable_empty_zones
-#   A hash containing a list of empty zones that shouldn't be created by bind
+#   An array containing a list of empty zones that shouldn't be created by bind
```
ekohl, line 138: The data type is already documented so perhaps we can simply reduce it to:
```suggestion
#   A list of empty zones that shouldn't be created by bind
```

PR #270, opened 2024-11-21
Fixes GH-269

### LINKED ISSUE #269: File logging channel enforces compulsory version and size
File logging channel enforces compulsory file_versions and file_size whereas it's not a compulsory field in bind configuration.

Ref: https://bind9.readthedocs.io/en/v9.18.14/reference.html#the-channel-phrase

https://github.com/theforeman/puppet-dns/blob/11.0.1/manifests/logging/channel.pp#L35

Some users are using log rotate to handle bind log rotation and don't want to specify file_versions and file_size.

Also file_versions also has an 'unlimited' option.

benjamin-robertson:
Fixes GH-269

I've made the decision not to allow users to specify a file_versions without a size. This configuration which will have no affect as the file will not rotate until it hits the file_size which by default is unlimited.

Unit tests passing. 

Fixes #269

### DIFF
--- a/spec/defines/dns_logging_channel_spec.rb
+++ b/spec/defines/dns_logging_channel_spec.rb
@@ -154,6 +154,17 @@ describe 'dns::logging::channel' do
         it { is_expected.not_to compile }
       end
 
+      describe 'with log_type = file and file_size and file_version not set' do
+        let(:params) do
+          {
+            'log_type'      => 'file',
+            'file_path'     => '/path/to/file.log',
+          }
+        end
+
+        it { is_expected.to compile.with_all_deps }
+      end
+
       describe 'with log_type = file and file_versions not set' do
         let(:params) do
           {
@@ -163,7 +174,7 @@ describe 'dns::logging::channel' do
           }
         end
 
-        it { is_expected.not_to compile }
+        it { is_expected.to compile.with_all_deps }
       end
 
       describe 'with log_type = syslog and all options set' do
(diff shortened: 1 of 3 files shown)


spec/defines/dns_logging_channel_spec.rb
```diff
@@ -163,7 +174,7 @@
           }
         end
 
-        it { is_expected.not_to compile }
+        it { is_expected.to compile.with_all_deps }
```
ekohl, line 177: I generally like to have more specific validation, something like
```suggestion
        it { is_expected.to compile.with_all_deps }
        it { is_expected.to contain_concat__fragment("named.conf-logging-channel-#{title}.dns").with_content(%r{file "/path/to/file.log" size 5m;});
```

spec/defines/dns_logging_channel_spec.rb
```diff
@@ -154,6 +154,17 @@
...
 
+      describe 'with log_type = file and file_size and file_version not set' do
+        let(:params) do
+          {
+            'log_type'      => 'file',
+            'file_path'     => '/path/to/file.log',
+          }
+        end
+
+        it { is_expected.to compile.with_all_deps }
```
ekohl, line 165: I generally like to have more specific validation, something like
```suggestion
        it { is_expected.to compile.with_all_deps }
        it { is_expected.to contain_concat__fragment("named.conf-logging-channel-#{title}.dns").with_content(%r{file "/path/to/file.log";});
```

ekohl (changes requested):
It took me a while to figure out exactly what it was doing, so I've modified the PR title to better cover it. Please modify it if you disagree.

benjamin-robertson:
Thanks for the review. The new title works for me and is a better description. I've accepted your suggestions and re-run the tests.

PR #271, opened 2025-01-12
fix path to named-checkconf on FreeBSD/DragonFly BSD

fraenki:
Fixes the following error:

```
Error: Execution of '/usr/local/sbin/named-checkconf /usr/local/etc/namedb/zones.conf20250112-57376-i5p19x' returned 1: Error: Could not execute posix command: No such file or directory - /usr/local/sbin/named-checkconf
```

Turns out the path to `named-checkconf` is incorrect:

```
$ uname -o
FreeBSD
$ which named-checkconf
/usr/local/bin/named-checkconf
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -69,7 +69,7 @@ class dns::params {
       $user               = 'bind'
       $group              = 'bind'
       $rndcconfgen        = '/usr/local/sbin/rndc-confgen'
-      $named_checkconf    = '/usr/local/sbin/named-checkconf'
+      $named_checkconf    = '/usr/local/bin/named-checkconf'
       # The sysconfig settings are not relevant for FreeBSD
       $sysconfig_file     = undef
       $sysconfig_template = undef
--- a/metadata.json
+++ b/metadata.json
@@ -69,7 +69,8 @@
     {
       "operatingsystem": "FreeBSD",
       "operatingsystemrelease": [
-        "11"
+        "13",
+        "14"
       ]
     },
     {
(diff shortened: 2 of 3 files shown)


fraenki:
The test failure does not seem to be related to this PR, the [previous test already failed](https://github.com/theforeman/puppet-dns/actions/runs/11669027531).

ekohl:
Do you happen to know if this change was specific to some FreeBSD version? Right now we only list FreeBSD 11 in our `metadata.json` which is rather ancient by now.

fraenki:
@ekohl It seems like this was changed with the introduction of the bind 9.18 port. The bind 9.16 still installed it into the `sbin` folder:
https://github.com/freebsd/freebsd-ports/blob/cfada54fb345c0886e2e11f0448bc0f7ed797299/dns/bind916/pkg-plist#L321
So it looks like this was introduced ~3 years ago during the FreeBSD 13.0 lifecycle.

fraenki:
Should I add FreeBSD 13 and 14 to metadata.json, and remove FreeBSD 11?

ekohl:
Yes please.

fraenki:
Done.

ekohl:
I wanted to change the PR summary to be about the OS version but realized we also only mention DragonFly BSD. Can we safely update that to 6 and does it include bind 9.18? I get the impression they used to ship bind but no longer do, but perhaps I'm just missing something.

fraenki:
I don't have a DragonFly machine at hand, but they definitely use the same path since about ~3 years:
https://github.com/DragonFlyBSD/DPorts/blob/master/dns/bind918/pkg-plist#L2

So I'd say: let's bump the DragonFly version to 6, too.

ekohl:
Ah, I couldn't find their ports but that looks correct. Let's bump it to 6 then.

ekohl:
> So I'd say: let's bump the DragonFly version to 6, too.

Mind updating the PR?

ikonia:
included and fixed in #302

PR #273, opened 2025-02-09
Add parameter to control listen-on

kajinamit:
... so that both listen-on and listen-on-v6 can be managed by the native parameters.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -43,6 +43,8 @@
 #   The forward option
 # @param forwarders
 #   The forwarders option
+# @param listen_on
+#   The listen-on option
 # @param listen_on_v6
 #   The listen-on-v6 option
 # @param recursion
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -159,6 +161,7 @@
   Variant[Enum['unmanaged'], Stdlib::Absolutepath] $defaultzonepath = $dns::params::defaultzonepath,
   Optional[Enum['only', 'first']] $forward                          = undef,
   Array[Dns::Forwarder] $forwarders                                 = [],
+  Optional[Variant[String, Boolean]] $listen_on                     = undef,
```
ekohl, line 164: Why a boolean instead of just an optional string?
```suggestion
  Optional[String] $listen_on                                       = undef,
```

Is it for consistency with `$listen_on_v6`? I'm not sure we need to care about that too much.

kajinamit: It is for consistency with listen_on_v6 but I agree we can probably ignore it. Let me update the change.

ekohl (approved):

PR #277, opened 2025-08-08
Add RPZ Zone Support

ikonia:
Add basic support to be able to define a response-policy list of zones into options.conf 
This will enable the value of https://github.com/theforeman/puppet-dns/pull/252 (a little rework maybe needed to align variable names) there is little point having the ability of response policies in views if you cannot define a response policy in the configuration.

This is also the foundation of being able to add more advanced scope of response policies in line with 
https://bind9.readthedocs.io/en/v9.20.11/reference.html#namedconf-statement-response-policy section 8.2.13.16

This work was done by also updating the options_conf.erb template to epp format in line with other previous changes and new features which have been delivered by epp templates rather than existing erb templates

config generation tested on EL8 EL9 EL10 with OpenVox 8 agent


ikonia:
config now also tested on Ubuntu 24.04 and Debian 12.11 (with 13 pending)

manifests/view.pp
```diff
@@ -52,7 +52,7 @@
   Optional[Enum['only','first']]        $forward              = undef,
   Optional[Enum['yes','no']]            $recursion            = undef,
   Optional[Enum['yes','no']]            $dnssec_enable        = undef,
-  Optional[Enum['yes','no']]            $dnssec_validation    = undef,
+  Optional[Enum['yes','no','auto']]       $dnssec_validation    = undef,
```
evgeni:
```suggestion
  Optional[Enum['yes','no','auto']]     $dnssec_validation    = undef,
```

ikonia:
obsoleted by pr #284

PR #278, opened 2025-08-26
Add support for Debian 13

smortex:
Debian 13 removed a bunch of default zones in favor of BIND 9 native
directive `empty-zones yes` (that is on by default).

### DIFF
--- a/spec/classes/dns_init_spec.rb
+++ b/spec/classes/dns_init_spec.rb
@@ -70,7 +70,9 @@ describe 'dns' do
       let(:localzonepath) do
         case facts[:os]['family']
         when 'Debian'
-          "#{etc_directory}/zones.rfc1918"
+          if facts[:os]['release']['major'] != '13'
+            "#{etc_directory}/zones.rfc1918"
+          end
         when 'RedHat'
           "#{etc_directory}/named.rfc1912.zones"
         end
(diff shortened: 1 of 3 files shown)


ikonia:
just incorporated this into my test branch too so I can test my change in Deb 13, works great.

smortex:
For the record, I also had to tweak `dnssec_validation` and set it to "auto" as the module default value is "[yes](https://github.com/theforeman/puppet-dns/blob/ed5dc308a20d45c82743e3ebab3d61a1bde8a845/manifests/init.pp#L172)".  According to the doc, "auto" is the default when not specified.  Not sure if we should reflect this in the module as it is a backwards-incompatible change.

ikonia:
@smortex the dnssec validation has been merged in now, if you rebased you'll find the tests will pass and your change will be ready to merge, I tested it against my branch already

ikonia:
looks good, looking forward to Debian 13 support

spec/classes/dns_init_spec.rb
ikonia: consideration is it worth doing > 13 rather than != as that way it will support Debian 14 onwards with no changes needed ?

ikonia (commented):
consideration only moving great than 13 as a validation for future version support

spec/classes/dns_init_spec.rb
ekohl: I'm ok with `!=` for now. Otherwise you also need to cover Ubuntu and I'd rather wait for that

ekohl (approved):

spec/classes/dns_init_spec.rb
smortex: > Otherwise you also need to cover Ubuntu and I'd rather wait for that

Yeah, IIRC it was the reason I did it that way.  I guess that this will be changed to a list of OS releases and a check to see if the current OS is part of that list, but as there is a single case for now I did it the easy way.

PR #279, opened 2025-09-23
Support CentOS Stream 10

kajinamit:


ekohl (commented):
I don't see the OS show up in acceptance tests so I think we need to update some things. Likely switch to OpenVox because that has EL10 builds

ekohl (approved):
Debian failure is unrelated
```
 {"errorDetail":{"message":"toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"},"error":"toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit"}
```

PR #284, opened 2025-10-16
enable rpz zone support

ikonia:
Add basic support to be able to define a response-policy list of zones into options.conf
This will enable the value of https://github.com/theforeman/puppet-dns/pull/252 (a little rework maybe needed to align variable names) there is little point having the ability of response policies in views if you cannot define a response policy in the configuration.

This is also the foundation of being able to add more advanced scope of response policies in line with
https://bind9.readthedocs.io/en/v9.20.11/reference.html#namedconf-statement-response-policy section 8.2.13.16

This work was done by also updating the options_conf.erb template to epp format in line with other previous changes and new features which have been delivered by epp templates rather than existing erb templates

config generation tested on EL8 EL9 EL10 with OpenVox 8 agent


templates/options.conf.epp
```diff
@@ -0,0 +1,51 @@
+directory "<%= $dns::vardir %>";
+<% if $dns::forwarders and $dns::forwarders != [] { -%>
+forwarders { <%= join($dns::forwarders, '; ') %>; };
+<% } -%>
+<% if $dns::forward != undef and $dns::forward != '' { -%>
+forward <%= $dns::forward %>;
+<% } -%>
+
+<% if $dns::rpz_zones { -%>
```
bastelfreak:
```suggestion
<% unless empty($dns::rpz_zones) { -%>
```

manifests/init.pp
```diff
@@ -161,6 +163,7 @@
   Variant[Enum['unmanaged'], Stdlib::Absolutepath] $defaultzonepath = $dns::params::defaultzonepath,
   Optional[Enum['only', 'first']] $forward                          = undef,
   Array[Dns::Forwarder] $forwarders                                 = [],
+  Optional[Array] $rpz_zones                                        = undef,
```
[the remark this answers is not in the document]
ikonia: a good lesson, acknowledged

templates/options.conf.epp
ikonia: happy to make this change, is there a reason unless empty is a better approach ?

templates/options.conf.epp
```diff
@@ -0,0 +1,51 @@
...
+recursion <%= $dns::recursion %>;
+allow-query { <%= join($dns::allow_query, '; ') %>; };
+<% if $dns::dnssec_enable != undef and $dns::dnssec_enable != '' and $dns::dnssec_enable != false { -%>
+dnssec-enable <%= $dns::dnssec_enable %>;
+<% } -%>
+dnssec-validation <%= $dns::dnssec_validation %>;
+
+empty-zones-enable <%= $dns::empty_zones_enable %>;
+
+<% if $dns::dns_notify != undef and $dns::dns_notify != '' { -%>
```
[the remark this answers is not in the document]
ikonia: need to look at this one, as this isn't something I changed in the logic, I just ported the current method to epp, I'll review and see what's happened here. 

Thanks you, helpful

ikonia:
improvements made aligned to review feedback, 
noted handling  listen_on_v6 could be potentially handled better

ikonia:
investigating why undef array is failing tests

manifests/init.pp
```diff
@@ -161,6 +163,7 @@
   Variant[Enum['unmanaged'], Stdlib::Absolutepath] $defaultzonepath = $dns::params::defaultzonepath,
   Optional[Enum['only', 'first']] $forward                          = undef,
   Array[Dns::Forwarder] $forwarders                                 = [],
+  Array $rpz_zones                                                  = undef,
```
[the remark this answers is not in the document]
ikonia: so that's interesting, I've missunderstood this, I thought 'undef' was always undefined no matter the datatype, and [] would be 'empty' rather than undefined.

templates/options.conf.epp
```diff
@@ -0,0 +1,51 @@
+directory "<%= $dns::vardir %>";
+<% if $dns::forwarders and $dns::forwarders != [] { -%>
+forwarders { <%= join($dns::forwarders, '; ') %>; };
+<% } -%>
+<% if $dns::forward != undef and $dns::forward != '' { -%>
```
bastelfreak: The variable cannot be an empty string an defaults to undef:
```suggestion
<% if $dns::forward { -%>
```

ikonia: missed in review, fixed with thanks

ikonia:
closing and re-opening merge requests to re-execute tests which appear to be false positive

ikonia:
tests pass, false failure was false positive with what looked like network comms problems

manifests/init.pp
```diff
@@ -161,6 +163,7 @@
   Variant[Enum['unmanaged'], Stdlib::Absolutepath] $defaultzonepath = $dns::params::defaultzonepath,
   Optional[Enum['only', 'first']] $forward                          = undef,
   Array[Dns::Forwarder] $forwarders                                 = [],
+  Array $rpz_zones                                                  = [],
```
bastelfreak: what's the data in the array? Always an Fqdn? Or we could use Stdlib::Fqdn?
```suggestion
  Array[Stdlib::Fqdn] $rpz_zones                                    = [],
```

https://github.com/puppetlabs/puppetlabs-stdlib/blob/main/types/fqdn.pp

bastelfreak: Can you please add a unit test that sets the parameter and then validate the config file?

ikonia: took a little while to test stdlib::fqdn as while in principal it will always be a fully qualified domain, a lot of orgs use non-standard domains for their RPZ zones, for security, blocking, etc, eg: org.security or logging.local.internal which thankfully do align to stlib::fqdn, I wasn't sure it would allow this but the pattern matching is quite simple.

I'll add this in and look at a usable test.

templates/options.conf.epp
```diff
@@ -0,0 +1,51 @@
+directory "<%= $dns::vardir %>";
+<% if $dns::forwarders and $dns::forwarders != [] { -%>
```
bastelfreak: it's always an array. So we can do:
```suggestion
<% unless empty($dns::forwarders) { -%>
```

ikonia: well spotted, and updated. thank you

manifests/init.pp
ikonia: think I've got a valid basic test in place now

bastelfreak (approved):

ikonia:
I've got a few more updates and enhancements I'd like to make to this module, but they depend on this PR being merged in so I can use the new EPP template anything stopping this getting merged now ? even if it's not released yet

PR #286, opened 2025-12-17
Fix permission of zonefilepath

kajinamit:
Add x bit because it's a directory.


ekohl (approved):
Technically I think Puppet already ensures this if you use `ensure => directory` (so it can also be correct when using recursive syncing with permissions) but it doesn't hurt to be explicit.

PR #290, opened 2026-01-19
Omit allow-recursion if empty

kajinamit:
Fix the regression introduced by 04301a6a , which causes invalid allow-recursion option when dns::allow_recursion is empty.


ekohl (approved):
Looks correct. Previous template also used `empty` and the data type is an array so this should be good.

PR #295, opened 2026-05-24
allow ACLS to be used in zone configuration options

ikonia:
--- suggested approach drop if a bad idea ----

being able to use ACLS' in zone definition such as allow-updates, allow-query, allow-tranfer is limited to internally defined ACLS (localnets and localhost) this maybe enough for a lot of use cases, but as ACL's and views become more important the ability to use ACLS within these parameters is important.

in order to use the ACL definition within puppet-dns the ACL's must be move higher up the template so they are available to be used in the zone definition.

this creates another problem that the validate process validates against temporary files from concat which means it will never pass validation because the required files it depends on are not generated at this point, until there is a better way to validate the ACL's it seems logical to remove validation on ACLS'

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -8,9 +8,9 @@ class dns::config {
   }
 
   concat { $dns::publicviewpath:
-    owner        => root,
-    group        => $dns::params::group,
-    mode         => '0640',
+    owner => root,
+    group => $dns::params::group,
+    mode  => '0640',
   }
 
   if $dns::enable_views {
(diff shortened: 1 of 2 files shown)


ikonia:
I believe this will also fix #164

ikonia:
could I get a review on this please ? I'm looking to put another one or two PR's in the next day or two that will depend on if this is going to be accepted, the PR's should make a reasonable functionality bump that warrants a release.

ikonia:
need a review on if disabling the validation on this function is 'ok' or if there is a better way to validate it

PR #296, opened 2026-05-24
add optional query-ipv6 parameter

ikonia:
more DNS servers are dual hosting on IPV6 and IPV4 certain conditions where an valid IPV6 address is not available on a bind server this causes a huge amount of noisy logs as ipv6 is consistently attempted (even with listen-on-ipv6 set to none) the answer in bind 9.18 is so use query-ipv6 parameter which will not lookup ipv6 records for name servers.
This is optional parameter so not breaking and will not be configured yes/no unless explicitly set, so is backward compatible and optional if you want to use it or not.

### DIFF
--- a/templates/options.conf.epp
+++ b/templates/options.conf.epp
@@ -32,7 +32,6 @@ listen-on { <%= $dns::listen_on %>; };
 <% if $dns::listen_on_v6 { -%>
 listen-on-v6 { <%= $dns::listen_on_v6 %>; };
 <% } -%>
-
 <% if $dns::query_ipv6 { -%>
 query-ipv6 <%= $dns::query_ipv6 %>;
 <% } -%>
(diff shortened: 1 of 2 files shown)


templates/options.conf.epp
```diff
@@ -33,6 +33,10 @@ listen-on { <%= $dns::listen_on %>; };
 listen-on-v6 { <%= $dns::listen_on_v6 %>; };
 <% } -%>
```
ekohl: I'd like to avoid empty lines for those who don't set the parameter. Perhaps group them together here?
```suggestion
```

ikonia: ok, I  see what you're saying, because options.conf was so heavy on spaces due to the conditions, I just didn't consider it. I can tidy up, nice feedback

ekohl (approved):

PR #299, opened 2026-08-20
modulesync 2026-08-20

bastelfreak:


ekohl:
Bleh, still uses `verify_contents` that we got rid of.

PR #301, opened 2026-08-29
Revert "Add optional query-ipv6 parameter"

ikonia:
TLDR: This reverts commit 065b550e378831e12d81c96d15a098f0c7149fe8.

Detail: 

Revert a feature that is present in bind 9.18 and later, initial testing of this feature was useful and worked, there was gaps in the implementation not fully considered. 

the feature was correctly introduced in https://gitlab.isc.org/isc-projects/bind9/-/merge_requests/9727 and while this was before EL10 release it was not earlier enough to make base that EL10 was built from.

There was no conditional verification of the feature implemented such as RHEL/Ubuntu/Version which in some situations could leave a broken feature in place (RHEL10)

A better approach is to revert this feature and remove the possibility (very edge case) of someone incorrectly using it, then re-implement it along with the raw-ipv6 parameter introduced in bind 9.19.19 as a pair with the correct verification of support and test cases associated with it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -51,8 +51,6 @@
 #   The listen-on option
 # @param listen_on_v6
 #   The listen-on-v6 option
-# @param query_ipv6
-#   The query-ipv6 option. Set to 'no' to disable IPv6 queries.
 # @param recursion
 #   The recursion option
 # @param allow_recursion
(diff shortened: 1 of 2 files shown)


ikonia:
re-reviewing if there is a way to build on what's there rather than revert and implement

ikonia:
closing and submitting an updated fix, better to fix forward and keep the functionality for the distros that support it

PR #302, opened 2026-08-30
Align module to currently supported releases and add Ubuntu 26.04

ikonia:
Add Ubuntu 26.04 support 

align other operating systems to currently supported operating system versions

FreeBSD is listed as 14 only, 13 has only just gone EOL, but I suspect this will cause no problems and is just a little aggressive on removing.

Questions on Ubuntu 20.04 - only in Ubuntu Pro support, is this something the module supports
Dragonfly is listed as 4 - what's the target, should dragon fly even be there any more ?

Tested manually on Ubuntu 26.04 Fedora 43 and 44 also.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -54,8 +54,8 @@
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
-        "39",
-        "40"
+        "43",
+        "44"
       ]
     },
     {
@@ -71,13 +71,14 @@
       "operatingsystemrelease": [
         "20.04",
         "22.04",
-        "24.04"
+        "24.04",
+        "26.04"
       ]
     },
     {
       "operatingsystem": "FreeBSD",
       "operatingsystemrelease": [
-        "11"
+        "14"
       ]
     },
     {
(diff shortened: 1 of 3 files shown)


ikonia:
investigating test failed on ubuntu 26.04 - tests passed when run locally as well as local test on an ubuntu 26.04 virtual machine

ikonia:
Tests fixed, not sure why they succeed locally when there was an error there, need to review how I'm executing tests locally

metadata.json
```diff
@@ -54,8 +54,8 @@
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
-        "39",
-        "40"
+        "43",
```
bastelfreak, line 57: I don't know the foreman support policy. IMO this is a breaking change and should be done in a dedicated PR)

ikonia: foreman doesn't support fedora at all, the module does support on it's own though, I don't see this as a breaking change as 43 or earlier is EOL anyway

ikonia:
included changes listed in #271 and version bumped dragon fly to 6 as per that PR, this should obsolete #271 at the same time that was waiting on author

metadata.json
ikonia: sorry that should read 42 or earlier, 43 is still in support, that's why it's included

d1nuc0m (approved):

PR #303, opened 2026-08-31
support ipv6_query properly evaluating distro support of bind 9.20

ikonia:
this work is what should have been submitted for #296 while it technically works it was a sloppy implementation, and in certain situations would have broken bind config if enabled (not a real problem as it could just be disabled with the parameter).

This work correctly evaluates distributions that are compatible with current supported versions of bind (9.19 onwards - as 9.18 is EOL June 2026) 

This updates baselines functional drift from 9.20 onwards anchoring on https://gitlab.isc.org/isc-projects/bind9/-/merge_requests/9727 changes. 

9.19 while supported is mostly 9.18 functionality with only minor parameter drift and all latest major distribution releases supported by this module ship with 9.20.

Test cases have been updated to only test 9.20 functionality against these distributions and future ones leaving backward compatibility for pre 9.20 distributions that are still in support (EL 8-9-10 / Ubuntu 20.04 / 22.04 / 24.04) until they go EOL. 

This PR also updates the metadata.json of supported OS's to align to actually current distributions, ensuring tests run against the correct OS's.  This update has also been shipped as a stand alone PR https://github.com/theforeman/puppet-dns/pull/302 should this not be merged in, it will at least bring the module in line with current supported versions of the OS's it supports, if this update is merged, I will close https://github.com/theforeman/puppet-dns/pull/302 

There is logic in the version evaluation for EL11 - while this is not released yet (and is not included in the metadata.json) it made sense to include the logic for it to both allow for testing against the Alpha releases (which contain Bind 9.20) and have it ready for release when RHEL 11 actually is released without having to change the logic, enable the OS in metadata.json and it's ready to go.

Tests pass locally and the functionality is checked as ignored on EL10, but applied on Ubuntu 26.04 and Debian 13

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -52,7 +52,16 @@
 # @param listen_on_v6
 #   The listen-on-v6 option
 # @param query_ipv6
-#   The query-ipv6 option. Set to 'no' to disable IPv6 queries.
+#   The query-source-v6 option. Set to 'no' to disable IPv6 queries for upstream
+#   name server connections. This option is only available in BIND 9.20 or later.
+#   For older BIND versions, use `sysconfig_startup_options => '-4'` instead.
+#   WARNING: Using this parameter with BIND < 9.20 will cause configuration errors.
+# @param bind_9_20_compat
+#   Internal parameter that indicates whether the system has BIND 9.20 or later.
+#   This is automatically detected based on the operating system version. Do not
+#   override this parameter unless you know the BIND version differs from the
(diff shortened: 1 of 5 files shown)


ikonia:
Still logic failures due to the complexity of bind versions and back ports of function. reviewing the syntax for the template in more detail, but fix forward is still the approach rather than remove

ikonia:
I'll wait for #302 to be merged in and rebase it on top of that.  I'll leave the PR in draft until that's in place

PR #304, opened 2026-09-08
Preserve address match list order in named.conf

### LINKED ISSUE #272: address match list ordering not respected
The ordering of elements within an address match list (i.e., an ACL) matters. Please note the paragraph starting with "Order of insertion is significant." at https://bind9.readthedocs.io/en/latest/reference.html#id4

But the [named.conf template sorts the elements](https://github.com/theforeman/puppet-dns/blob/master/templates/named.conf.erb#L34), breaking many address match lists where order matters (and the array isn't coincidentally already in sort order).

bwitt:
Preserve address match list order in named.conf, since it's significant in BIND.

Fixes https://github.com/theforeman/puppet-dns/issues/272

### DIFF
--- a/spec/classes/dns_init_spec.rb
+++ b/spec/classes/dns_init_spec.rb
@@ -374,12 +374,44 @@ describe 'dns' do
         ])}
       end
 
+      describe 'with acls whose order must be preserved' do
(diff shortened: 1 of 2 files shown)


ikonia (commented):
functionality looks good.

